PR #3, opened 2020-09-01
+

dan3si:
[DEMO LINK](https://dan3si.github.io/react_goods-selector/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -14,11 +14,51 @@ const goodsFromServer = [
   'Garlic',
 ];
 
-const App = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
+    selectedGood: 'No selected goods',
+  }
+
+  selector = (good) => {
+    this.setState({ selectedGood: good });
+  }
+
+  render() {
+    return (
+      <div className="App">
+        <h1>
+          Selected good: -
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -14,11 +14,51 @@ const goodsFromServer = [
...
+          {` ${this.state.selectedGood}`}
+        </h1>
+
+        {goodsFromServer.map(good => (
+          <button
+            type="button"
+            className={good === this.state.selectedGood
+              ? 'good good--selected'
+              : 'good'
+            }
```
ArtemiiKab, line 40: wow, hacker)

ArtemiiKab (approved):

src/App.js
dan3si: А чё, можно было как-то по-другому? :D

> 
> 
> wow, hacker)

src/App.js
```diff
@@ -14,11 +14,51 @@ const goodsFromServer = [
...
+
+        <button
+          type="button"
+          className="cleaner"
+          onClick={() => {
+            this.setState({ selectedGood: 'No selected goods' });
+          }}
+        >
+          Clean
+        </button>
```
vkryvytskyy, line 58: show this button only when some good is selected

src/App.scss
```diff
@@ -1 +1,34 @@
-// styles go here
+.App {
```
vkryvytskyy, line 1: other classes starts with lowercase letter, use one naming convention

PR #8, opened 2020-09-02
Solution

uni-8:
- [DEMO LINK](https://uni-8.github.io/react_goods-selector/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -14,11 +14,87 @@ const goodsFromServer = [
   'Garlic',
 ];
 
-const App = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
+    selectedGood: '',
+  }
+
+  addSelection = (event, good) => {
+    if (!event.ctrlKey) {
+      this.setState({ selectedGood: good });
+
+      return;
+    }
+
+    event.stopPropagation();
+
+    if (this.state.selectedGood.includes(good)) {
+      this.setState(state => (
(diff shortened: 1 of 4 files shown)


KaterinaSavvova (approved):

src/App.js
```diff
@@ -14,11 +15,86 @@ const goodsFromServer = [
...
+  }
+
+  addSelection = (event, item) => {
+    if (!event.ctrlKey) {
+      this.setState({ selectedGood: item });
+
+      return;
+    }
+
+    event.stopPropagation();
```
vkryvytskyy, line 30: why do you use it?

src/App.js
```diff
@@ -14,11 +15,86 @@ const goodsFromServer = [
...
+
+    if (this.state.selectedGood.includes(item)) {
```
vkryvytskyy, line 32: use destructuring

src/App.js
```diff
@@ -14,11 +15,86 @@ const goodsFromServer = [
...
+
+        <p>
+          All goods:
+          {' '}
+          {goodsFromServer.length}
+        </p>
+
+        <ul className="App__list">
+          {goodsFromServer.map(item => (
+            <li key={Math.random()}>
```
vkryvytskyy, line 82: don't use random as key, use `item`(it's a unique string, i believe)

vkryvytskyy (changes requested):
image
looks strange in firefox

src/App.js
uni-8: i thought it was necessary for proper work

vkryvytskyy (approved):

PR #111, opened 2021-01-16
solution

Pasha5555:
[DEMO LINK](https://pasha5555github.io/react_goods-selector/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -14,11 +14,88 @@ const goodsFromServer = [
   'Garlic',
 ];
 
-const App = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
+    selected: [],
+  }
+
+  toAdd = (good) => {
+    if (!this.state.selected.includes(good)) {
+      this.setState(prevState => ({
+        selected: [...prevState.selected, good],
+      }));
+    }
+  }
+
+  toRemove = (good) => {
+    this.setState((prevState) => {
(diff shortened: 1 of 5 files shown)


zhenya-mezhueva (approved):

src/App.js
```diff
@@ -14,11 +14,88 @@ const goodsFromServer = [
...
+
+      return {
+        selected: prevState.selected,
+      };
+    });
+  }
+
+  clearSelection = () => {
+    this.setState({
+      selected: '',
```
varseniuk, line 46: 'selected' in your state is initially an array. Why it becomes a string here?
That's why we need propTypes (or typeScript). We always should stick to one data type

src/App.js
```diff
@@ -14,11 +14,88 @@ const goodsFromServer = [
...
+          className="button-clear buttons"
+          onClick={this.clearSelection}
+        >
+          Clear
+        </button>
+        <ul className="list">
+          {goodsFromServer.map(good => (
+            <li
+              key={good}
+              className={selectedArr.includes(good) ? 'selected' : ''}
```
varseniuk, line 71: It's a good case to use ClassNames library. Or 
```suggestion
              className={selectedArr.includes(good) && 'selected'}
```

varseniuk (changes requested):

src/App.js
Pasha5555: Консоль скаржиться що клас не заданий якщо юзати &&

varseniuk: And that's why we need ClassNames library

varseniuk (approved):
👍

PR #112, opened 2021-01-17
Develop

VinnieJ-2k20:
- [DEMO LINK](https://VinnieJ-2k20.github.io/react_goods-selector/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -14,11 +14,85 @@ const goodsFromServer = [
   'Garlic',
 ];
 
-const App = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
+    selectedGood: null,
+    header: null,
+  }
+
+  markSelectedGood = (event, product) => {
+    const { selectedGood } = this.state;
+    const item = event.target.closest('.list__item');
+
+    if (selectedGood) {
+      selectedGood.classList.remove('list__item--selected');
+    }
+
+    item.classList.add('list__item--selected');
+
+    this.setState({
+      selectedGood: item,
+      header: product,
+    });
+  }
+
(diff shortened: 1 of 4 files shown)


tamara-ostapets (approved):
great work 🌟

src/App.js
```diff
@@ -14,11 +14,102 @@ const goodsFromServer = [
...
+  state = {
+    selectedGoods: [],
+    products: [],
+  }
+
+  addOrRemoveSelectedGood = ({ target }, product) => {
+    this.setState(({ selectedGoods, products }) => {
+      const item = target.closest('.list__item');
+
+      item.classList.toggle('list__item--selected');
```
varseniuk, line 27: Let's not use DOM methods. Only React

src/App.js
```diff
@@ -14,11 +14,102 @@ const goodsFromServer = [
...
+        products: newProducts,
+      };
+    });
+  }
+
+  clearSelectedGoods = (event) => {
+    const { selectedGoods } = this.state;
+
+    selectedGoods.forEach((item) => {
+      item.classList.remove('list__item--selected');
```
varseniuk, line 53: Same here - don't use DOM methods

varseniuk (changes requested):

VinnieJ-2k20:
Indeed, turns out it all can be done in a much more simple way.

src/App.js
```diff
@@ -14,11 +15,88 @@ const goodsFromServer = [
...
+      selectedGoods: [],
+    });
+  }
+
+  printHeader = () => {
+    const { selectedGoods } = this.state;
+
+    if (selectedGoods.length === 0) {
+      return '';
+    }
```
varseniuk, line 50: You don't need this. If your array is empty (length === 0), it will just return nothing and the string will be empty

varseniuk (approved):
Good. Read my comment

PR #114, opened 2021-01-17
added solution

MarchD:
[DEMO LINK](https://MarchD.github.io/react_goods-selector/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -14,11 +14,89 @@ const goodsFromServer = [
   'Garlic',
 ];
 
-const App = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
+    basket: [],
+  }
+
+  toSelect = ({ target }) => {
+    target.parentElement.classList.add('selected');
+
+    const targetName = goodsFromServer.find((good) => {
+      const titleOfGood = target.parentElement.firstChild.textContent;
+
+      return good === titleOfGood;
+    });
+    const prevStateOfBasket = this.state.basket;
+
+    this.setState({ basket: [...prevStateOfBasket, targetName] });
+  };
+
+  toCancel = ({ target }) => {
+    const titleOfGood = target.parentElement.firstChild.textContent;
+    const index = this.state.basket.indexOf(titleOfGood);
+
+    this.state.basket.splice(index, 1);
+    const itemsOfBasket = this.state.basket;
+
+    itemsOfBasket.splice(1, index);
+    this.setState({ basket: [...itemsOfBasket] });
+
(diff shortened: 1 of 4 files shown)


dtsehelnyk (approved):
Здорово)
Держи ёлочку 🎄

src/App.js
```diff
@@ -14,11 +14,86 @@ const goodsFromServer = [
...
+class App extends React.Component {
+  state = {
+    basket: [],
+  }
+
+  toToggle = ({ target }) => {
+    const { basket } = this.state;
+    const titleOfGood = target.parentElement.firstChild.textContent;
+
+    target.parentElement.classList.toggle('selected');
```
vsm2k17, line 26: don't use DOM

src/App.js
```diff
@@ -14,11 +14,86 @@ const goodsFromServer = [
...
+    this.setState({ basket: [...basket, title] });
+  };
+
+  toCancel = (title) => {
+    const { basket } = this.state;
+    const index = basket.indexOf(title);
+    const itemsOfBasket = basket;
+
+    itemsOfBasket.splice(index, 1);
+    this.setState({ basket: [...itemsOfBasket] });
```
vsm2k17, line 45: if you want to remove an element from the array it's easier to use the filter method
but you don't need here array it should be only one selected item

src/App.js
```diff
@@ -14,11 +14,86 @@ const goodsFromServer = [
...
+  }
+
+  toReset = () => {
+    const goods = [...document.querySelectorAll('.good')];
```
vsm2k17, line 49: don't use DOM

src/App.js
```diff
@@ -14,11 +14,86 @@ const goodsFromServer = [
...
+
+    goods.forEach((good) => {
+      const button = good.querySelector('button');
+
+      good.classList.remove('selected');
```
vsm2k17, line 54: don't use classList

src/App.js
```diff
@@ -14,11 +14,86 @@ const goodsFromServer = [
...
+      button.innerText = 'Add';
```
vsm2k17, line 55: also no

src/App.js
```diff
@@ -14,11 +14,86 @@ const goodsFromServer = [
...
+            {this.state.basket.join(', ')}
+          </h1>
+        </div>
+
+        <div className="goods">
+          {goodsFromServer.map(good => (
+            <div key={good} className="good">
+              <p>{good}</p>
+              <button
+                className="button"
```
vsm2k17, line 85: you can add condition here and react set classes only if the condition is true

vsm2k17 (changes requested):
image
only one means that if click on another good it replaces the previous one

update: just ignore the comment above, I forgot about advanced tasks

vsm2k17 (approved):
good job

PR #116, opened 2021-01-18
develop

Layrovell:
[DEMO LINK](https://Layrovell.github.io/react_goods-selector/)

### DIFF
--- a/src/components/Selector/Selector.jsx
+++ b/src/components/Selector/Selector.jsx
@@ -2,15 +2,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Button } from 'react-bulma-components';
 
-class Selector extends React.Component {
+export class Selector extends React.Component {
   state = {
     selectedGoods: [],
-  };
-
-  deleteItem = (item) => {
-    this.setState(prevState => ({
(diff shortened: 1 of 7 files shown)


Serhii-Alieksieievych (approved):
не впевнений що так має бути, але в вимогах це прямо не забороняється, тому апруваю)
https://prnt.sc/x5kdpd

varseniuk (changes requested):
Serhii is right. You should add only one instance of any good at a time. Fix it. Everything else is ok

src/components/Selector/Selector.jsx
```diff
@@ -0,0 +1,92 @@
...
+export class Selector extends React.Component {
+  state = {
+    selectedGoods: [],
+  };
+
+  addItem = (item) => {
+    this.setState(prevState => ({
+      selectedGoods: [...prevState.selectedGoods, item],
+    }));
+  };
```
varseniuk, line 14: Your function still adds multiple same items to the list. You can prevent this by adding check here like
`if(!this.state.selectedGoods.includes(item)) {...}`

varseniuk (changes requested):
Still can add multiple items

varseniuk (approved):
👍

PR #117, opened 2021-01-18
first solution

troospieler:
[DEMO LINK](https://troospieler.github.io/react_goods-selector/)

array of items changed according to Denys Cheporniuk recommendation

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,24 +1,78 @@
+/* eslint-disable object-curly-newline */
 import React from 'react';
 import './App.scss';
+import classNames from 'classnames';
 
 const goodsFromServer = [
-  'Dumplings',
-  'Carrot',
-  'Eggs',
-  'Ice cream',
-  'Apple',
-  'Bread',
-  'Fish',
-  'Honey',
-  'Jam',
-  'Garlic',
+  { id: 1, name: 'Dumplings' },
+  { id: 2, name: 'Carrot' },
+  { id: 3, name: 'Eggs' },
+  { id: 4, name: 'Ice cream' },
+  { id: 5, name: 'Apple' },
+  { id: 6, name: 'Bread' },
+  { id: 7, name: 'Fish' },
+  { id: 8, name: 'Honey' },
+  { id: 9, name: 'Jam' },
+  { id: 10, name: 'Garlic' },
 ];
 
-const App = () => (
-  <div className="App">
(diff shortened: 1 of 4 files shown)


src/App.js
```diff
@@ -1,24 +1,78 @@
...
-  'Carrot',
-  'Eggs',
-  'Ice cream',
-  'Apple',
-  'Bread',
-  'Fish',
-  'Honey',
-  'Jam',
-  'Garlic',
+  { id: 1, name: 'Dumplings' },
```
tamara-ostapets, line 7: I suppose, since all goods are unique strings, you could use them as `key`

src/App.js
```diff
@@ -1,24 +1,78 @@
...
 ];
 
-const App = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
```
tamara-ostapets, line 20: you could create a `selectedGoods` property in the state (initially, an empty array) so there was an option to select multiple goods at the same time

tamara-ostapets (commented):
As I understood, button Select should select and deselect the product, if it's selected, at the same time

tamara-ostapets (approved):

src/App.js
```diff
@@ -1,24 +1,78 @@
...
+  render() {
+    const { selectedId } = this.state;
+    const showGood = goodsFromServer.find(good => good.id === selectedId);
+
+    return (
+      <div className="App">
+        <h1>
+          Selected good: -
+          {' '}
+          {showGood ? showGood.name : ''}
```
varseniuk, line 41:
```suggestion
          {showGood && showGood.name}
```

src/App.js
```diff
@@ -1,24 +1,78 @@
...
+    selectedId: null,
+  }
+
+  cleaner = () => {
+    this.setState({ selectedId: null });
+  }
+
+  selected = (id) => {
+    this.setState({ selectedId: id });
+  }
```
varseniuk, line 30:
```suggestion
  handleSelect = (id) => {
    this.setState({ selectedId: id });
  }
```
Let's give more meaningful names for such merhods

varseniuk (changes requested):
All is good, but let's make it better

varseniuk (approved):
Good 👍

PR #124, opened 2021-01-18
Solution

Iryna713:
[DEMO LINK](https://Iryna713.github.io/react_goods-selector/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import { Product } from './components/Product';
 
 const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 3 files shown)


chri911 (approved):

src/App.js
```diff
@@ -14,11 +15,79 @@ const goodsFromServer = [
...
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
+    selected: [],
+  }
+
+  clear = () => {
+    const goodsList = document.querySelector('.goods');
```
vsm2k17, line 24: don't use document and DOM

src/App.js
```diff
@@ -14,11 +15,79 @@ const goodsFromServer = [
...
+
+    [...goodsList.children].forEach((good) => {
+      good.classList.remove('active');
```
vsm2k17, line 27: don`t use classList

src/App.js
```diff
@@ -14,11 +15,79 @@ const goodsFromServer = [
...
+    });
+
+    this.setState({ selected: [] });
+  }
+
+  add = ({ target }) => {
+    const listTarget = target.closest('li');
+    const newGood = listTarget.firstChild.textContent;
+
+    listTarget.classList.toggle('active');
```
vsm2k17, line 37: don't use target when you can(here you can avoid this) and classList, add a condition into className and add class active when you need

src/App.js
```diff
@@ -14,11 +15,79 @@ const goodsFromServer = [
...
+
+    if (listTarget.classList.contains('active')) {
```
vsm2k17, line 39: remove classList

vsm2k17 (changes requested):
don't use DOM

src/App.js
```diff
@@ -14,11 +15,71 @@ const goodsFromServer = [
...
+class App extends React.Component {
+  state = {
+    selected: [],
+  }
+
+  clear = () => {
+    this.setState({ selected: [] });
+  }
+
+  add = (good) => {
```
vsm2k17, line 27: you call the method "add" but it can more than that

vsm2k17 (approved):
good job, but use better names

PR #130, opened 2021-01-18
Completed

Serhii-Alieksieievych:
[DEMO LINK](https://Serhii-Alieksieievych.github.io/react_goods-selector/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -14,11 +14,108 @@ const goodsFromServer = [
   'Garlic',
 ];
 
-const App = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+let selectedGoodsArr = [];
 
-export default App;
+export class App extends React.Component {
+  state = {
+    selectedGoodName: '',
+  }
+
+  separator = ' ----- ';
+
+  render() {
+    return (
+      <div className="App">
+        <h1>
+          Selected good:
+          {this.state.selectedGoodName}
+        </h1>
+
+        <button
+          type="button"
+          onClick={(event) => {
+            selectedGoodsArr = [];
+            this.setState(state => ({
+              selectedGoodName: selectedGoodsArr
+                .map(good => good
+                  .innerText
+                  .split(this.separator)[0])
+                .join(', '),
+            }));
+            const items = [...document.body.querySelectorAll('.selected')];
+
+            items.forEach(item => item.classList.remove('selected'));
+          }}
+        >
+          Clear
+        </button>
+
+        <ul>
+          {goodsFromServer.map(good => (
+            <li
+              key={good}
+              className={this.state.itemClass}
+            >
+              {good}
+              {this.separator}
+              <button
+                type="button"
+                onClick={(event) => {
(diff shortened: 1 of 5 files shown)


Layrovell (approved):

src/App.js
```diff
@@ -14,11 +14,108 @@ const goodsFromServer = [
...
+          onClick={(event) => {
+            selectedGoodsArr = [];
+            this.setState(state => ({
+              selectedGoodName: selectedGoodsArr
+                .map(good => good
+                  .innerText
+                  .split(this.separator)[0])
+                .join(', '),
+            }));
+            const items = [...document.body.querySelectorAll('.selected')];
```
vsm2k17, line 45: don't use DOM

vsm2k17: use condition inside className for classes and use the state to save all data that you need

vsm2k17 (changes requested):
remove usage of DOM

src/App.js
```diff
@@ -14,11 +14,108 @@ const goodsFromServer = [
...
-    {goodsFromServer.length}
-  </div>
-);
-
-export default App;
+export class App extends React.Component {
+  state = {
+    goods: [...goodsFromServer],
+    selectedGoods: [],
+    selectedGoodsNames: [],
```
vsm2k17, line 21: should be only one array in your state

src/App.js
```diff
@@ -14,11 +14,108 @@ const goodsFromServer = [
...
+                `${selectedGoodsNames.includes(good) ? 'selected' : ''}`}
+            >
+              {good}
+
+              {' '}
+
+              <button
+                type="button"
+                onClick={(event) => {
+                  const selectedGood = event.target.parentElement;
```
vsm2k17, line 65: don't use event.target.parentElement, just pass the name (good) in your callback

src/App.js
```diff
@@ -14,11 +14,108 @@ const goodsFromServer = [
...
+                  const textOfSelectedGood = event.target
```
vsm2k17, line 66: textOfSelectedGood === good in your array goods, so you can use it without DOM

src/App.js
```diff
@@ -14,11 +14,108 @@ const goodsFromServer = [
...
+                  }
+
+                  this.setState(state => ({
+                    selectedGoods: [
+                      ...state.selectedGoods,
+                      selectedGood,
+                    ],
+                    selectedGoodsNames: [
+                      ...state.selectedGoodsNames,
+                      textOfSelectedGood,
```
vsm2k17, line 81: git rid of selectedGoodsNames

src/App.js
```diff
@@ -14,11 +14,108 @@ const goodsFromServer = [
...
+
+              {' '}
+
+              <button
+                type="button"
+                onClick={(event) => {
+                  const selectedGood = event.target.parentElement;
+                  const textOfSelectedGood = event.target
+                    .parentElement.childNodes[0]
+                    .wholeText.trim();
```
vsm2k17, line 97: just delete an element from an array which you map to render the list (without state mutation)

vsm2k17 (changes requested):

PR #133, opened 2021-01-19
Solution

sasha39612:
[DEMO LINK](https://sasha39612.github.io/react_goods-selector/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -14,11 +14,87 @@ const goodsFromServer = [
   'Garlic',
 ];
 
-const App = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+const preparedGoods = goodsFromServer.map((item, index) => (
+  {
+    id: index,
+    name: item,
+  }
+));
+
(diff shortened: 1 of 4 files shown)


Serhii-Alieksieievych (approved):

src/App.js
```diff
@@ -14,11 +14,87 @@ const goodsFromServer = [
...
+ clearSelect = () => {
+   this.setState({
+     selectedGoods: [],
+   });
+ }
+
+ render() {
+   const { selectedGoods } = this.state;
+
+   const temp = ((selectedGoods.length > 0) ? 1 : 0);
```
varseniuk, line 56: don't you think it would be better to make it boolean? true/false === 1/0
```suggestion
   const temp = Boolean(selectedGoods.length);
```
and better to give it a more meaningful name. Like `areGoodsSelected` or smth like this

varseniuk (approved):
Good solution, but let's make it cleaner

src/App.js
sasha39612: I did't know it. Thank's.
 I'm going to use it in future.

PR #134, opened 2021-01-19
Solution

olesiatk:


Serhii-Alieksieievych (approved):

src/App.js
```diff
@@ -14,11 +15,84 @@ const goodsFromServer = [
...
+  state = {
+    selectedGoods: [],
+  };
+
+  selectElement = ({ target }) => {
+    const newGood = target.parentElement.firstChild.textContent;
+    const goods = this.state.selectedGoods;
+
+    target.parentElement.classList.toggle('selected');
+    this.setState({ selectedGoods: [...goods, newGood] });
```
vsm2k17, line 28: don't use target and DOM, if you want to change class, then add condition into className attribute, if you need any kind of data store them into the state, and use it from the state.

vsm2k17 (changes requested):
remove usage of DOM

src/App.js
olesiatk: Fixed

src/App.js
```diff
@@ -14,11 +15,80 @@ const goodsFromServer = [
...
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
+    selectedGoods: [],
+  };
+
+  selectElement = (good) => {
+    const goods = this.state.selectedGoods;
```
vsm2k17, line 24:
```suggestion
    const { selectedGoods: goods } = this.state;
```

src/App.js
```diff
@@ -14,11 +15,80 @@ const goodsFromServer = [
...
+
+  resetGoods = () => {
+    this.setState({ selectedGoods: [] });
+  }
+
+  render() {
+    const selected = this.state.selectedGoods;
+    const select = this.selectElement;
+    const remove = this.removeElement;
+    const reset = this.resetGoods;
```
vsm2k17, line 47:
```suggestion
    const { selectedGoods: selected } = this.state;
    const { select, remove, reset } = this;
```

vsm2k17 (approved):
good job

PR #233, opened 2021-05-24
first solution

scarlettjohanssons:
[DEMO LINK](https://scarlettjohanssons.github.io/react_goods-selector/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,9 +1,10 @@
 import React from 'react';
+import classNames from 'classnames';
 import './App.scss';
 
 const goodsFromServer = [
   'Dumplings',
-  'Carrot',
+  'Cgoodsot',
   'Eggs',
   'Ice cream',
   'Apple',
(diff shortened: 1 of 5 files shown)


rauzok (approved):
nice!

src/App.js
```diff
@@ -14,11 +15,98 @@ const goodsFromServer = [
...
+            <div className="goods__item" key={good}>
+              <span>
+                {good}
+              </span>
+
+              <button
+                type="button"
+                className={
+                  classNames('btn', {
+                    good__hidden: selectedGoods.includes(good),
```
mykola-yarchuk, line 74: use conditional rendering to hide button

src/App.js
```diff
@@ -14,11 +15,98 @@ const goodsFromServer = [
...
+                }}
+              >
+                Add
+              </button>
+
+              <button
+                type="button"
+                className={
+                  classNames({
+                    good__hidden: !selectedGoods.includes(good),
```
mykola-yarchuk, line 90: conditional rendering

src/App.js
```diff
@@ -14,11 +15,98 @@ const goodsFromServer = [
...
+                    good__remove: true,
```
mykola-yarchuk, line 91: if you always need some class, just add it outside of the object

mykola-yarchuk, line 91:
```suggestion
                  classNames('good__remove', {
                    good__hidden: !selectedGoods.includes(good),
```

src/App.js
```diff
@@ -14,11 +15,98 @@ const goodsFromServer = [
...
+                  })
+                }
+                onClick={() => {
+                  this.setState({
+                    selectedGoods: selectedGoods.filter(
+                      product => product !== good,
+                    ),
+                  });
```
mykola-yarchuk, line 99: is better to create a method in advance and pass it here as callback

mykola-yarchuk (changes requested):

src/App.js
```diff
@@ -14,11 +14,95 @@ const goodsFromServer = [
...
+
+              default: {
+                return `${selectedGoods
+                  .slice(0, (selectedGoods.length - 1))
+                  .join(', ')}`
+                + ` and ${selectedGoods[selectedGoods.length - 1]}`
+                + ` are selected`;
+              }
+            }
+          })()}
```
mykola-yarchuk, line 58: is better to create a function and use switch inside of it

src/App.js
```diff
@@ -14,11 +14,95 @@ const goodsFromServer = [
...
+          {goodsFromServer.map(good => (
+            <div className="goods__item" key={good}>
+              <span>
+                {good}
+              </span>
+
+              <button
+                type="button"
+                className={
+                  selectedGoods.includes(good) ? 'good__hidden' : 'btn'
```
mykola-yarchuk, line 80: you can create some variable (`isIncluded` for example) inside callback in `map` method before you return `<li>`, and use this variable in all places where you need it. with this approach, you can use the method includes only once

mykola-yarchuk: use conditional rendering to hide something

src/App.js
```diff
@@ -14,11 +14,95 @@ const goodsFromServer = [
...
+                  });
+                }}
+              >
+                Add
+              </button>
+
+              <button
+                type="button"
+                className={
+                  selectedGoods.includes(good) ? 'good__remove' : 'good__hidden'
```
mykola-yarchuk, line 94: conditional rendering

mykola-yarchuk (changes requested):

mykola-yarchuk (approved):

PR #239, opened 2021-05-24
Solution

herkoss:
https://herkoss.github.io/react_goods-selector/

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import classNames from 'classnames';
 import './App.scss';
 
 const goodsFromServer = [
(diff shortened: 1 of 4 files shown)


src/App.js
```diff
@@ -14,11 +15,91 @@ const goodsFromServer = [
...
+            </button>
+          )}
+        </h1>
+        <span>{goodsFromServer.length}</span>
+        <ul>
+          {goodsFromServer.map(good => (
+            <li
+              key={good}
+              className={classNames({
+                selected: this.state.selectedGoods.includes(good),
```
mykola-yarchuk, line 75: you can create some variable (`isIncluded` for example) inside callback in `map` method before you return `<li>`, and use this variable in all places where you need it. with this approach, you can use the method includes only once

src/App.js
```diff
@@ -14,11 +15,91 @@ const goodsFromServer = [
...
+
+  selectGood = good => this.setState(prevState => ({
+    selectedGoods: [...prevState.selectedGoods, good],
+  }));
+
+  deleteGood = good => (
+    this.setState((prevState) => {
+      const allGoods = [...prevState.selectedGoods];
+
+      allGoods.splice(allGoods.indexOf(good), 1);
```
mykola-yarchuk, line 31: filter is better for this case

mykola-yarchuk (changes requested):

mykola-yarchuk (approved):

PR #244, opened 2021-05-24
Solution-001

senkiv-oleh:
[DEMO LINK](https://senkiv-oleh.github.io/react_goods-selector/)


illisogori (approved):
Good job.

src/App.js
```diff
@@ -14,11 +15,96 @@ const goodsFromServer = [
...
-  </div>
-);
+class App extends React.Component {
+  state = {
+    selectedGood: ['Jam'],
+  };
+
+  componentDidUpdate() {
+    document.querySelectorAll('li').forEach((element) => {
+      if (this.state.selectedGood.includes(element.textContent)) {
```
varseniuk, line 25: Let's not use DOM methods while studying react. You can use 'classanmes' library for conditional styles applying

src/App.js
```diff
@@ -14,11 +15,96 @@ const goodsFromServer = [
...
+      } else {
+        element.classList.remove('hover');
+      }
+    });
+  }
+
+  selectedGoods = (selectedGood) => {
+    switch (selectedGood.length) {
+      case 0: return `No goods selected`;
+      case 1: return `${selectedGood} is selected`;
```
varseniuk, line 44:
```suggestion
      case 0:
        return `No goods selected`;
      case 1:
        return `${selectedGood} is selected`;
```
For better readability

src/App.js
```diff
@@ -14,11 +15,96 @@ const goodsFromServer = [
...
+    document.querySelectorAll('.clearAll-button').forEach((element) => {
+      if (this.state.selectedGood.length === 0) {
+        element.classList.add('hover');
+      } else {
+        element.classList.remove('hover');
+      }
+    });
+  }
+
+  selectedGoods = (selectedGood) => {
```
varseniuk, line 41:
```suggestion
  getSelectedGoodsTitle = (selectedGood) => {
```
Let's use more meaningful names. It's better to name the functions according to what they actually do

src/App.js
```diff
@@ -14,11 +15,96 @@ const goodsFromServer = [
...
+            {this.selectedGoods(selectedGood)}
+          </h1>
+          <button
+            title="Clear the list of selected itemss"
+            type="button"
+            className="clearAll-button"
+            onClick={() => {
+              this.setState(() => ({
+                selectedGood: [],
+              }));
```
varseniuk, line 70:
```suggestion
              this.setState({ selectedGood: [] });
```
You can use simple setState (not functional) here because there's no need in previous state values

src/App.js
```diff
@@ -14,11 +15,96 @@ const goodsFromServer = [
...
+          >
+            X
+          </button>
+        </>
+
+        <ul className="goods">
+          {goodsFromServer.map(good => (
+            <div className="goods__wrapper" key={good}>
+              <li
+                className={`goods__item ${good} ${good === 'Jam' ? 'active' : ''}`}
```
varseniuk, line 81: Use 'classnames' library here
Also, why do you need `good` class? (`${good}`)

varseniuk (changes requested):
Good start!
But let's get rid of classList.toggle (DOM methods) and use 'classnames' library instead

varseniuk (approved):
Great job! 👍

PR #252, opened 2021-05-25
Solution

Taras-Konovalov:
[DEMO LINK](https://Taras-Konovalov.github.io/react_goods-selector/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.scss';
+import classNames from 'classnames';
 
 const goodsFromServer = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


src/App.js
```diff
@@ -14,11 +15,91 @@ const goodsFromServer = [
...
+          >
+            X
+          </button>
+        </div>
+        <ul className="Product__list">
+          {goodsFromServer.map(product => (
+            <li
+              key={product}
+              className={classNames('Product__item', {
+                Product__item_added: selectedGoods.includes(product),
```
mykola-yarchuk, line 75: you can create some variable (`isIncluded` for example) inside callback in `map` method before you return `<li>`, and use this variable in all places where you need it. with this approach, you can use the method includes only once

mykola-yarchuk (changes requested):

mykola-yarchuk (approved):

PR #255, opened 2021-05-25
Solution

SergeyZub:
[DEMO LINK](https://SergeyZub.github.io/react_goods-selector/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,24 +1,70 @@
 import React from 'react';
 import './App.scss';
 
-const goodsFromServer = [
-  'Dumplings',
-  'Carrot',
-  'Eggs',
-  'Ice cream',
-  'Apple',
-  'Bread',
-  'Fish',
-  'Honey',
-  'Jam',
-  'Garlic',
-];
+class App extends React.Component {
+  state = {
+    goodsFromServer: [
+      'Dumplings',
+      'Carrot',
+      'Eggs',
+      'Ice cream',
+      'Apple',
+      'Bread',
+      'Fish',
+      'Honey',
+      'Jam',
+      'Garlic',
+    ],
+    selectedGood: 'Jam',
+  }
 
-const App = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+  render() {
+    const { goodsFromServer, selectedGood } = this.state;
+
+    return (
+      <div className="App">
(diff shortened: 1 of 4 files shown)


progmari (approved):
👍

src/App.js
```diff
@@ -1,24 +1,70 @@
...
-  'Apple',
-  'Bread',
-  'Fish',
-  'Honey',
-  'Jam',
-  'Garlic',
-];
+class App extends React.Component {
+  state = {
+    goodsFromServer: [
```
varseniuk, line 6:
```suggestion
  state = {
    goods: [...goodsFromServer]
```
No need to write the whole array in the state. You could create a copy for this

src/App.js
```diff
@@ -1,24 +1,70 @@
...
+        {selectedGood ? (
+          <h1>
+            {selectedGood}
+            {' '}
+            is selected
+            <button
+              type="button"
+              onClick={() => {
+                this.setState({
+                  selectedGood: null,
```
varseniuk, line 35: Better to create a separate function for this (smth like `clearAllGoods`) and use a callback here onClick

src/App.js
```diff
@@ -1,24 +1,70 @@
...
+          </h1>
+        ) : (
+          <h1>No goods selected</h1>
+        )}
+        {goodsFromServer.map(good => (
+          <div className={selectedGood === good ? 'active' : 'good'} key={good}>
+            {good}
+            {selectedGood === good ? (''
+            ) : (
+              <button
```
varseniuk, line 50:
```suggestion
            {selectedGood !== good ?? (
              <button
```
Let's use proper conditional rendering

varseniuk (changes requested):
Good job! But let's use proper conditional rendering (without ternary and empty strings) and it's done

src/App.js
```diff
@@ -14,11 +14,60 @@ const goodsFromServer = [
...
+              onClick={this.clearAllGoods}
+            >
+              x
+            </button>
+          </h1>
+        ) : (
+          <h1>No goods selected</h1>
+        )}
+        {goods.map(good => (
+          <div className={selectedGood === good ? 'active' : 'good'} key={good}>
```
varseniuk, line 50: One more moment - it's better to use `ul`-`li` when rendering lists

varseniuk (approved):
Good!
Also, in future use `ul` when rendering lists

PR #265, opened 2021-05-25
React goods selector

Allrightman:
[DEMO LINK](https://Allrightman.github.io/react_goods-selector/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -14,11 +14,53 @@ const goodsFromServer = [
   'Garlic',
 ];
 
-const App = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
+    selected: 'Jam',
+  };
+
+  render() {
+    const { selected } = this.state;
+
+    return (
+      <div className="App">
+        {selected ? (
+          <h1>
+            {`${selected} is selected`}
+            <button
+              className="cancel"
+              type="button"
+              onClick={() => {
+                this.setState({ selected: null });
+              }}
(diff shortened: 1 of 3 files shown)


src/App.js
```diff
@@ -14,11 +14,53 @@ const goodsFromServer = [
...
+              onClick={() => {
+                this.setState({ selected: null });
+              }}
+            >
+              X
+            </button>
+          </h1>
+        ) : (
+          <h1>No goods selected</h1>
+        )}
```
mykola-yarchuk, line 42:
```suggestion
        {selected 
          ? (
          <h1>
            {`${selected} is selected`}
            <button
              className="cancel"
              type="button"
              onClick={() => {
                this.setState({ selected: null });
              }}
            >
              X
            </button>
          </h1>
        ) 
        : (
          <h1>No goods selected</h1>
        )}
```

src/App.js
```diff
@@ -14,11 +14,53 @@ const goodsFromServer = [
...
+        {selected ? (
+          <h1>
+            {`${selected} is selected`}
+            <button
+              className="cancel"
+              type="button"
+              onClick={() => {
+                this.setState({ selected: null });
+              }}
+            >
```
mykola-yarchuk, line 36: is better to create a method and use it as a callback

mykola-yarchuk (approved):

src/App.js
Allrightman: Ok, this one is short so I put it inside, but yeah - callbacks looks better

src/App.js
Allrightman: Thanks a lot!

PR #267, opened 2021-05-25
terrible solution

dmitry-puhliakov:
[DEMO LINK](https://dmitry-puhliakov.github.io/react_goods-selector/)


varseniuk (approved):
Solution is great 👍 
Don't know why you think it's terrible)

PR #268, opened 2021-05-25
test

C-y-y-y:
[DEMO LINK](https://C-y-y-y.github.io/react_goods-selector/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -14,11 +14,51 @@ const goodsFromServer = [
   'Garlic',
 ];
 
-const App = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
+    selectedGood: 'Jam',
+  };
+
+  render() {
+    const { selectedGood } = this.state;
+
+    return (
+      <>
+        {selectedGood ? (
+          <h1>
+            {`Selected good: - ${selectedGood} is selected`}
+            <button
+              type="button"
+              onClick={() => {
+                this.setState({ selectedGood: null });
+              }}
+            >
+              x
+            </button>
+          </h1>
+        ) : (
+          <h1>No value selected</h1>
+        )}
+
+        {goodsFromServer.map(n => (
+          <li type="square">
+            <button
+              type="button"
+              key={n}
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -14,11 +14,51 @@ const goodsFromServer = [
...
+              }}
+            >
+              x
+            </button>
+          </h1>
+        ) : (
+          <h1>No value selected</h1>
+        )}
+
+        {goodsFromServer.map(n => (
```
varseniuk, line 43:
```suggestion
        {goodsFromServer.map(good => (
```
Let's give more meaningful names

src/App.js
```diff
@@ -14,11 +14,51 @@ const goodsFromServer = [
...
+          <li type="square">
+            <button
+              type="button"
+              key={n}
+              className={selectedGood === n ? 'active' : ''}
```
varseniuk, line 48: In the future for such purposes better to use 'classnames' library

src/App.js
```diff
@@ -14,11 +14,51 @@ const goodsFromServer = [
...
+            >
+              x
+            </button>
+          </h1>
+        ) : (
+          <h1>No value selected</h1>
+        )}
+
+        {goodsFromServer.map(n => (
+          <li type="square">
```
varseniuk, line 44: `li` tag without `ul` is not valid.
Render all your list items inside `ul`

varseniuk (changes requested):
Good job! But let's not use `li` outside of `ul`. Fix this and it's done

src/App.js
```diff
@@ -14,11 +16,53 @@ const goodsFromServer = [
...
+              x
+            </button>
+          </h1>
+        ) : (
+          <h1>No value selected</h1>
+        )}
+
+        {goodsFromServer.map(good => (
+          <ul>
+            <li type="square">
```
varseniuk, line 47:
```suggestion
        <ul>
          {goodsFromServer.map(good => (
            <li key={good} type="square">
```
Now you are creating many lists `ul` with one list item `li`. You should create one list with many mapped items inside.
Also, the `key` attribute should be applied to `li`, not to `button`

varseniuk (changes requested):
Let's fix list rendering

varseniuk (approved):

PR #269, opened 2021-05-25
so;ution

MinchukO:
DEMO LINK](https://minchuko.github.io/react_goods-selector/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -14,11 +14,107 @@ const goodsFromServer = [
   'Garlic',
 ];
 
-const App = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
+    title: 'No goods selected',
+    // eslint-disable-next-line react/no-unused-state
+    list: goodsFromServer,
+    selectedGood: [],
+  };
+
+  onAddProduct = (newTitle) => {
+    this.state.selectedGood.push(newTitle);
+    this.setState({
+      title: `${newTitle} is selected`,
+    });
+  };
+
+  onDeleteProduct = (product) => {
+    const index = this.state.selectedGood.indexOf(product);
+
+    if (index > -1) {
+      const { selectedGood } = this.state;
+      const arr = selectedGood.splice(index, 1);
+
+      this.setState({
+        selectedGood: arr,
+        title: `${product} is deleted`,
+      });
+    } else {
+      this.setState({
+        title: `${product} was not selected`,
+      });
+    }
+  };
+
+  clear = () => {
+    this.setState({
+      selectedGood: ['Jam'],
+      title: 'No goods selected',
+    });
+  };
+
+  render() {
+    return (
+      <div className="wrapper">
+        <h1>{this.state.title}</h1>
+        <button type="button" onClick={this.clear}>
+          X
+        </button>
+        <p>
+          <strong>Your list: </strong>
+          {this.state.selectedGood.join(', ')}
+        </p>
+        <div
+          style={{
+            width: 700,
+            border: '1px solid #ccc',
+            backgroundColor: 'beige',
+            margin: '20px auto',
+          }}
+        >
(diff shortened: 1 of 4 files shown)


src/App.js
```diff
@@ -14,11 +14,105 @@ const goodsFromServer = [
...
+      selectedGood: [],
+      title: 'No goods selected',
+    });
+  };
+
+  render() {
+    return (
+      <div className="wrapper">
+        <h1>{this.state.title}</h1>
+        <button type="button" onClick={this.clear}>X</button>
```
Namart, line 62: кнопка должна появляться только, если что-то выбрано

src/App.js
```diff
@@ -14,11 +14,105 @@ const goodsFromServer = [
...
+        >
+          {this.state.list.map(product => (
+            <>
+              <ul className="list" key={product}>
+                <li
+                  className={
+                    this.state.selectedGood.includes(product)
+                      ? 'list--bg'
+                      : null
+                  }
```
Namart, line 83: тут можно использовать && оператор вместо тернарного

src/App.js
```diff
@@ -14,11 +14,105 @@ const goodsFromServer = [
...
+                  {product}
+                </li>
+                <button
+                  type="button"
+                  onClick={this.onDeleteProduct.bind(this, product)}
+                  className={
+                    !this.state.selectedGood.includes(product)
+                      ? 'visibility'
+                      : null
+                  }
```
Namart, line 94: тут тоже &&

src/App.js
```diff
@@ -14,11 +14,105 @@ const goodsFromServer = [
...
+                >
+                  Delete
+                </button>
+                <button
+                  type="button"
+                  onClick={this.onAddProduct.bind(this, product)}
+                  className={
+                    this.state.selectedGood.includes(product)
```
Namart, line 102: и тут)

Namart (changes requested):

src/App.js
```diff
@@ -14,11 +14,111 @@ const goodsFromServer = [
...
+            width: 700,
+            border: '1px solid #ccc',
+            backgroundColor: 'beige',
+            margin: '20px auto',
+          }}
+        >
+          {this.state.list.map(product => (
+            <>
+              <ul className="list" key={product}>
+                <li
```
varseniuk, line 84:
```suggestion
          <ul className="list">
            {this.state.list.map(product => (
              <li key={product}
```
You should render (map) list items inside `ul` tag. Now you have a lot of lists (`ul`). There should be one `ul` with many `li` inside.
Also, `key` attribute should be applied to `li` (not `ul`)

src/App.js
varseniuk: Better to use 'classname' library

src/App.js
varseniuk: same

varseniuk (changes requested):
Let's fix some issues. Especially you should pay attention to state mutation and list rendering.
And move styles to css

src/App.js
```diff
@@ -14,11 +14,96 @@ const goodsFromServer = [
...
+    });
+  };
+
+  onDeleteProduct = (product) => {
+    const index = this.state.selectedGood.indexOf(product);
+
+    if (index > -1) {
+      const { selectedGood: prev } = this.state;
+
+      prev.splice(index, 1);
```
varseniuk, line 39:
```suggestion
      prev.splice(index, 1);
```
better to use `filter` method for deleting

src/App.js
```diff
@@ -14,11 +14,96 @@ const goodsFromServer = [
...
+          X
+        </button>
+        <p>
+          <strong>Your list: </strong>
+          {this.state.selectedGood.join(', ')}
+        </p>
+        <div>
+          <ul className="list">
+            {this.state.list.map(product => (
+              <>
```
varseniuk, line 73:
```suggestion
```
No need in fragment here

varseniuk (approved):
Good job!
Left some comments

PR #272, opened 2021-05-25
created goods-selector

YuriiLiso:
[DEMO LINK](https://yuriiliso.github.io/react_goods-selector/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -14,11 +14,58 @@ const goodsFromServer = [
   'Garlic',
 ];
 
-const App = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
+    selected: 'Jam',
+  }
+
+  render() {
+    const { selected } = this.state;
+
+    return (
+      <>
+        {selected
+          ? (
+            <h1>
+              Selected good: -
+              {selected}
+
+              <button
+                type="button"
+                onClick={() => {
+                  this.setState({ selected: null });
+                }}
+              >
+                X
+              </button>
+            </h1>
+          ) : (
+            <h1>
+              No goods selected
+            </h1>
+          )}
+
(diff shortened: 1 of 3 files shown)


taya294 (approved):
👍

src/App.js
```diff
@@ -14,11 +14,58 @@ const goodsFromServer = [
...
+              </button>
+            </h1>
+          ) : (
+            <h1>
+              No goods selected
+            </h1>
+          )}
+
+        {goodsFromServer.map(el => (
+          <>
```
varseniuk, line 49:
```suggestion
```
No need in this fragment here

src/App.js
```diff
@@ -14,11 +14,58 @@ const goodsFromServer = [
...
+            <div>
+              {el}
+              {this.state.selected !== el && (
+                <button
+                  type="button"
+                  key={el}
```
varseniuk, line 55: `key` should be on the `div` that rendered on the top level.
Actualy, it's better to use `ul`-`li` structure here. Semantically it's a list

varseniuk (changes requested):
Great start!
Let's use `ul-li` instead of `div` and it's done

src/App.js
```diff
@@ -14,11 +14,57 @@ const goodsFromServer = [
...
+            </h1>
+          ) : (
+            <h1>
+              No goods selected
+            </h1>
+          )}
+
+        {goodsFromServer.map(el => (
+          <ul key={el}>
+            <li>
```
varseniuk, line 50:
```suggestion
        <ul>
          {goodsFromServer.map(el => (
            <li key={el}>
```
Now you are creating a lot of lists with one item. It's wrong. You should create one list (`ul`) with many list items (`li`). Also, key shoul be used for `li` not `ul`

varseniuk (changes requested):
Let's fix issue with list rendering

varseniuk (approved):
Good!

PR #273, opened 2021-05-25
Dev

den-bulaev:
[DEMO LINK](https://den-bulaev.github.io/react_goods-selector/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -14,11 +14,51 @@ const goodsFromServer = [
   'Garlic',
 ];
 
-const App = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
+    good: 'No goods selected',
+  }
+
+  render() {
+    const { good } = this.state;
+
+    return (
+      <div className="App">
+        <h1>{`Selected good: ${good}`}</h1>
+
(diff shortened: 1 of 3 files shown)


src/App.js
```diff
@@ -14,11 +14,51 @@ const goodsFromServer = [
...
+        >
+          clear
+        </button>
+
+        <br />
+        <br />
+
+        <ul>
+          {goodsFromServer.map(item => (
+            <li className={good.includes(item) ? 'yellow' : ''}>
```
mykola-yarchuk, line 42: you can create some variable (`isIncluded` for example) inside callback in `map` method before you return `<li>`, and use this variable in all places where you need it. with this approach, you can use the method includes only once

src/App.js
```diff
@@ -14,11 +14,51 @@ const goodsFromServer = [
...
+
+  render() {
+    const { good } = this.state;
+
+    return (
+      <div className="App">
+        <h1>{`Selected good: ${good} is selected`}</h1>
+
+        <button
+          className={good === 'Nothing is selected' ? 'selectedButton' : ''}
```
mykola-yarchuk, line 30: hide this button with conditional rendering

mykola-yarchuk (changes requested):

mykola-yarchuk (approved):

PR #274, opened 2021-05-25
solution

tanyakovchuk:
[DEMO LINK](https://taya294.github.io/react_goods-selector/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -16,25 +16,25 @@ const goodsFromServer = [
 
 class App extends React.Component {
   state = {
-    values: [],
+    value: null,
   }
 
   render() {
-    const { values } = this.state;
+    const { value } = this.state;
 
     return (
       <div className="App">
         <h1>
-          { values.length > 0
+          { value
             ? (
               <>
                 Selected good:
                 {' '}
-                {values}
+                {value}
                 <button
                   type="button"
                   onClick={() => {
-                    this.setState({ values: [] });
+                    this.setState({ value: null });
                   }}
                 >
                   x
@@ -49,13 +49,13 @@ class App extends React.Component {
           { goodsFromServer.map(item => (
             <li key={item} className="list__item">
               <span className="list__text">{item}</span>
-              {values.includes(item)
+              {value === item
                 ? (
                   <button
                     type="button"
-                    className={values === item ? 'active' : ''}
+                    className={value === item ? 'active' : ''}
                     onClick={() => {
-                      this.setState({ values: item });
+                      this.setState({ value: item });
                     }}
                   >
                     +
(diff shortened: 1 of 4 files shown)


src/App.js
```diff
@@ -14,11 +14,66 @@ const goodsFromServer = [
...
+            <li className="list__item">
+              <span className="list__text">{item}</span>
+              {values.includes(item)
+                ? (
+                  <button
+                    type="button"
+                    key={item}
+                    className={values === item ? 'active' : ''}
+                    onClick={() => {
+                      this.setState({ values: item });
```
YuriiLiso, line 53: You just rewrite ```values```, and I guess you want to add to the array.

YuriiLiso (changes requested):

YuriiLiso (approved):
God job!

src/App.js
```diff
@@ -14,11 +14,70 @@ const goodsFromServer = [
...
 
-const App = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
+    values: [],
```
varseniuk, line 19:
```suggestion
    value: null,
```
In fact, you have only one value selected. no need to create an array

src/App.js
```diff
@@ -14,11 +14,70 @@ const goodsFromServer = [
...
+  }
+
+  render() {
+    const { values } = this.state;
+
+    return (
+      <div className="App">
+        <h1>
+          { values.length > 0
+            ? (
```
varseniuk, line 29:
```suggestion
          { value ? (
```
If you will have `value` instead of `values` you may use it like this

src/App.js
```diff
@@ -14,11 +14,70 @@ const goodsFromServer = [
...
+              <>
+                Selected good:
+                {' '}
+                {values}
+                <button
+                  type="button"
+                  onClick={() => {
+                    this.setState({ values: [] });
```
varseniuk, line 37:
```suggestion
                    this.setState({ value: null });
```
Again - no need to create an array if it's only one value

src/App.js
```diff
@@ -14,11 +14,70 @@ const goodsFromServer = [
...
+        <ul>
+          {' '}
+          { goodsFromServer.map(item => (
+            <li key={item} className="list__item">
+              <span className="list__text">{item}</span>
+              {values.includes(item)
+                ? (
+                  <button
+                    type="button"
+                    className={values === item ? 'active' : ''}
```
varseniuk, line 56: For switching classes, it's better to use 'classnames' library (not ternary operator with empty strings)
Not obligatory for this task, but use it in the future

src/App.js
```diff
@@ -14,11 +14,70 @@ const goodsFromServer = [
...
+                      this.setState({ values: item });
+                    }}
+                  >
+                    +
+                  </button>
+                )
+                : (
+                  <button
+                    type="button"
+                    className={values === item ? 'active' : ''}
```
varseniuk, line 67: Same here

varseniuk (changes requested):
Good job! 👍 
Let's just change unnecessary array to a string and it's done

varseniuk (approved):
Great job! 🚀

PR #338, opened 2021-09-01
Created goods selector

Anna-Tkachenko:
[DEMO LINK](https://Anna-Tkachenko.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import classNames from 'classnames';
 import './App.scss';
 
 const goodsFromServer: string[] = [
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -14,11 +15,93 @@ const goodsFromServer: string[] = [
...
+      ? `Selected goods: ${stringOfGoods}`
+      : 'No goods selected';
+
+    return (
+      <div className="App">
+        <div className="App__header">
+          <h1>
+            {listTitle}
+          </h1>
+          {selectedGoods.length !== 0 && (
```
varseniuk, line 57:
```suggestion
          {!!selectedGoods.length && (
```

varseniuk (approved):
🔥 
Left some suggestions

P.S: would be great to implement the additional points from the task (but only if you have some time):
```
4. All the selected goods should be listed in the h1:
- No goods selected
- Jam is selected
- Carrot, Jam and Eggs are selected
```

PR #343, opened 2021-09-02
solution react-goods-selector 

AndreySerhovetc:
[DEMO LINK](https://AndreySerhovetc.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import classNames from 'classnames';
 import './App.scss';
 
 const goodsFromServer: string[] = [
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -14,11 +15,100 @@ const goodsFromServer: string[] = [
...
+          ? (
+            <button
+              className="card__btn card__btn--reset"
+              type="button"
+              onClick={resetState}
+            >
+              Clear
+            </button>
+          )
+          : null}
```
ihor-jpeg, line 74: in this piece of code you are checking if the array is empty or not. you want to simply a button or nothing, this looks like place to use logical operator "and". 

let's say `selectedGoods.length` is 0; you can cast it to boolean using a double "not" operator `!!`.

`!!selectedGoods.length` will equal to false, if the length is 0. this way you can write:

`!!selectedGoods.length && (
              <button
                className="card__btn card__btn--reset"
                type="button"
                onClick={resetState}
              >
                Clear
              </button>
            )`

src/App.tsx
```diff
@@ -14,11 +15,100 @@ const goodsFromServer: string[] = [
...
+        <div className="container">
+          {goodsFromServer.map(good => (
+            <div className={classNames(
+              'card',
+              { card__active: selectedGoods.includes(good) },
```
ihor-jpeg, line 79: active is a modifier, i think it's better to write it like `card--active`

src/App.tsx
```diff
@@ -14,11 +15,100 @@ const goodsFromServer: string[] = [
...
+            )}
+            >
+              <h2 className="card__title" key={good}>
```
ihor-jpeg, line 82: if i am not mistaken, key attribute should go on the first child element

ihor-jpeg (changes requested):
i think you are on the right way! i left a few suggestions below. 

In the task description it is required that after adding an item to the cart, the add button should disappear, because this way you can add the same item multiple times. also it doesn't make sense to show and be able to click the remove button if the item is not yet in the cart

src/App.tsx
AndreySerhovetc: fixed)

src/App.tsx
AndreySerhovetc: fixed)

src/App.tsx
AndreySerhovetc: fixed)

src/App.tsx
```diff
@@ -14,11 +15,100 @@ const goodsFromServer: string[] = [
...
+    selectedGoods: [],
+  };
+
+  addToSelectedGood = (good: string) => {
+    this.setState((state) => ({
+      selectedGoods: [...state.selectedGoods, good],
+    }));
+  };
+
+  removeToSelectedGood = (good: string) => {
```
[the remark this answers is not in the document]
AndreySerhovetc: fixed)

src/App.tsx
```diff
@@ -14,11 +15,100 @@ const goodsFromServer: string[] = [
...
+    const { selectedGoods } = this.state;
+
+    if (selectedGoods.includes(good)) {
+      const index = selectedGoods.lastIndexOf(good);
+
+      selectedGoods.splice(index, 1);
```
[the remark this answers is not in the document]
AndreySerhovetc: fixed

SashaRudenko:
Андрей, отличная работа! Но функционал реализован немного не до конца. В строке отображения товаров должно отображаться 'are selected' или' is selected' в зависимости от количества. А также 'and' перед последним товаром.

SashaRudenko (approved):

src/App.tsx
```diff
@@ -14,11 +15,90 @@ const goodsFromServer: string[] = [
...
+  };
+
+  addToSelectedGood = (good: string) => {
+    this.setState((state) => ({
+      selectedGoods: [...state.selectedGoods, good],
+    }));
+  };
+
+  removeFromSelectedGood = (good: string) => {
+    this.setState((state) => ({
```
Denchepornyuk, line 34: тут аналогічно, назву краще змінити щоб не виникало проблем з неймінгом, бо змінна state вже оголошена

```suggestion
    this.setState((currentState) => ({
```

Denchepornyuk (changes requested):
Screenshot 2021-09-06 at 21 01 21
а що тут робить цей ноль?

src/App.tsx
AndreySerhovetc: fixed

Denchepornyuk (approved):

PR #344, opened 2021-09-02
Goods selector

nastyyvolk:
[DEMO LINK](https://nastyyvolk.github.io/react_goods-selector/)


src/App.tsx
```diff
@@ -14,11 +15,97 @@ const goodsFromServer: string[] = [
...
+    selectedGoods: [],
+  };
+
+  addProduct = (product: string) => {
+    this.setState((prevState: State): State => {
+      const { selectedGoods } = prevState;
+
+      return { selectedGoods: [...selectedGoods, product] };
+    });
+  };
```
ihor-jpeg, line 33: here you could have simply use this.state. are we required to do this bit `(prevState: State): State`?
i'm genuinely curious

src/App.tsx
```diff
@@ -14,11 +15,97 @@ const goodsFromServer: string[] = [
...
+      if (i === 0) {
+        return item;
+      }
+
+      if (i !== length - 1) {
+        return `${prev}, ${item}`;
+      }
+
+      return `${prev} and ${item}`;
+    }, '');
```
ihor-jpeg, line 60: a unique solution, i'd say

src/App.tsx
```diff
@@ -14,11 +15,97 @@ const goodsFromServer: string[] = [
...
+
+    return (
+      <div className="app">
+        <h1 className="app__heading">
+          {length > 0 ? `${whatIsSelected} ${toBeForm} selected` : 'Nothing is selected'}
+        </h1>
+
+        {selectedGoods.length > 0 && (
```
ihor-jpeg, line 68: you can turn any value to it's boolean equivalent by using `!!` operator. not that it will reduce the amount of code or anything, but i'll just leave it here

ihor-jpeg (approved):
🔥 great work! 
i left a few comments below.

src/App.tsx
nastyyvolk: for some reason, linter wanted me to do that yesterday.
today, it works without `: State`

src/App.tsx
nastyyvolk: the idea to use `!! ` just didn't cross my mind

Denchepornyuk: так якщо буде !!0 то це буде false і тобі не потрібно буде робити перевірки на 0
Якщо довжина буде більше чим 1 то !!1 поверне true

Denchepornyuk (approved):

Denchepornyuk (approved):

PR #376, opened 2021-09-08
Develop

RomSmile:
DEMO LINK : https://RomSmile.github.io/react_goods-selector/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,24 +1,79 @@
 import React from 'react';
+// import { render } from 'react-dom';
 import './App.scss';
 
-const goodsFromServer: string[] = [
-  'Dumplings',
-  'Carrot',
-  'Eggs',
-  'Ice cream',
-  'Apple',
-  'Bread',
-  'Fish',
-  'Honey',
-  'Jam',
-  'Garlic',
-];
+type State = {
+  goodsFromServer: string[],
+  selectedGood: string[],
+};
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component<{}, State> {
+  state: State = {
+    goodsFromServer: [
+      'Dumplings',
+      'Carrot',
+      'Eggs',
+      'Ice cream',
+      'Apple',
+      'Bread',
+      'Fish',
+      'Honey',
+      'Jam',
+      'Garlic'],
+    selectedGood: [],
+  };
+
+  addItem = (item: string) => {
+    this.setState((state) => ({
+      selectedGood: [...state.selectedGood, item],
+    }));
+  };
+
+  removeItem = (item: string) => {
+    this.setState((state) => ({
+      selectedGood: state.selectedGood.filter(x => x !== item),
+    }));
+  };
+
+  clearGoods = () => {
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,24 +1,79 @@
...
+      'Dumplings',
+      'Carrot',
+      'Eggs',
+      'Ice cream',
+      'Apple',
+      'Bread',
+      'Fish',
+      'Honey',
+      'Jam',
+      'Garlic'],
```
ihor-jpeg, line 22: i don't think this the way to do it. you will be receiving data from server, and it doesn't have to be in your state, as that list does not change at all.

src/App.tsx
```diff
@@ -1,24 +1,79 @@
...
+        <div className="App">
+          <h1>
+            Selected good:
+            <p className="goods">
+              {selectedGood.length > 0
+                ? <button type="button" className="clearButton" onClick={this.clearGoods}>X</button>
+                : ''}
+              {selectedGood.join(',')}
+            </p>
+            {}
```
ihor-jpeg, line 56: and this is?

src/App.tsx
```diff
@@ -1,24 +1,79 @@
...
+
+          </h1>
+
+          <ul>
+            {goodsFromServer.map((good) => (
+              <li
+                key={good}
+                className={selectedGood.find(x => x === good) ? 'selected' : 'notSelected'}
```
ihor-jpeg, line 64: i have to give you props for this one. i did not think of it and had to use the classnames library. well done!

ihor-jpeg (approved):
great work! 👍  i left a few comment below

src/App.tsx
```diff
@@ -1,24 +1,79 @@
...
-  </div>
-);
+type State = {
+  goodsFromServer: string[],
+  selectedGood: string[],
+};
+
+class App extends React.Component<{}, State> {
+  state: State = {
+    goodsFromServer: [
```
Denchepornyuk, line 12: весь масив краще винести за межі компоненти App

src/App.tsx
```diff
@@ -1,24 +1,79 @@
...
+      'Carrot',
+      'Eggs',
+      'Ice cream',
+      'Apple',
+      'Bread',
+      'Fish',
+      'Honey',
+      'Jam',
+      'Garlic'],
+    selectedGood: [],
```
Denchepornyuk, line 23: це завдання можна зробити і з одним масивом в state

src/App.tsx
```diff
@@ -1,24 +1,79 @@
...
+  };
+
+  addItem = (item: string) => {
+    this.setState((state) => ({
```
Denchepornyuk, line 27:
```suggestion
    this.setState((currentState) => ({
```

src/App.tsx
Denchepornyuk, line 56: шо тут робить пустий об'єкт

src/App.tsx
Denchepornyuk, line 64: замість find тут можна використати includes

Denchepornyuk (changes requested):
Screenshot 2021-09-09 at 11 18 37
jam має бути по дефолту виділений

danheim (approved):

PR #415, opened 2021-10-14
solution

VladSaen:
- [DEMO LINK](https://VladSaen.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -56,11 +56,11 @@ class App extends React.Component<{}, State> {
           : ''}
 
         {goodsFromServer.map(item => (
-          <label
+          <div
             key={item}
             className={selectedGood.includes(item)
-              ? 'label label--selected'
-              : 'label'}
+              ? 'goods goods--selected'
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -14,11 +14,76 @@ const goodsFromServer: string[] = [
...
+  state = {
+    selectedGood: ['Jam'],
+  };
+
+  selected = (goods: string[]) => {
+    if (goods.length === 1) {
+      return ` ${goods[0]}`;
+    }
+
+    return goods.map(item => (goods.indexOf(item) + 1 !== goods.length
```
Anna-Tkachenko, line 31: You can use index from arguments of callback function
```suggestion
    return goods.map((item, index) => index + 1 !== goods.length
```

Anna-Tkachenko (approved):

src/App.tsx
VladSaen: right, I forgot

PR #427, opened 2021-10-15
Solution

DY-JS:
[DEMO LINK](https://DY-JS.github.io/react_goods-selector/)


Artem5457 (approved):

src/App.tsx
```diff
@@ -14,11 +14,98 @@ const goodsFromServer: string[] = [
...
+  state = {
+    selectedGoods: ['Jam'],
+  };
+
+  selectHandler = (item: string) => {
+    const { selectedGoods } = this.state;
+    const copy = [...selectedGoods, item];
+
+    this.setState({ selectedGoods: copy });
+  };
```
aliceblunt8, line 31: also, it would be nice to use functional setState
https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

```suggestion
  handleSelect = (item: string) => {
    this.setState((prevState) => ({ selectedGoods: [...prevState.selectedGoods, item] }));
  };
```

src/App.tsx
```diff
@@ -14,11 +14,98 @@ const goodsFromServer: string[] = [
...
+
+  removeSelectHandler = (item: string) => {
```
aliceblunt8, line 33:
```suggestion
  handleRemoval = (item: string) => {
```

src/App.tsx
```diff
@@ -14,11 +14,98 @@ const goodsFromServer: string[] = [
...
+    const { selectedGoods } = this.state;
+    const index = selectedGoods.indexOf(item);
+    const copy = [...selectedGoods];
+
+    copy.splice(index, 1);
```
aliceblunt8, line 38: maybe pass good's name and filter array instead of indexOf and splice?

src/App.tsx
```diff
@@ -14,11 +14,98 @@ const goodsFromServer: string[] = [
...
+          className={selectedGoods.length > 0
+            ? 'app__clear' : 'app__clear--none'}
+          onClick={this.clearList}
+        >
+          ❌ Clear List
+        </button>
+
+        {goodsFromServer.map((item) => {
+          const flag = selectedGoods.includes(item);
+          const cls = flag ? 'app__item app__item--selected' : 'app__item';
```
aliceblunt8, line 87: you can use classnames library for conditional classes

src/App.tsx
```diff
@@ -14,11 +14,98 @@ const goodsFromServer: string[] = [
...
+                className="app__btn"
+                onClick={() => {
+                  if (flag) {
+                    this.removeSelectHandler(item);
+                  } else {
+                    this.selectHandler(item);
+                  }
+                }}
+                value={flag ? 'Remove' : 'Add'}
+              />
```
aliceblunt8, line 103: it's better to render 2 different buttons instead adding if else inside onClick

PR #460, opened 2021-11-11
Goods selector

Anton-Karpena:
[DEMO LINK](https://Anton-Karpena.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,14 @@
 import React from 'react';
 import './App.scss';
+import { v4 as uuidv4 } from 'uuid';
+import classNames from 'classnames';
 
-const goodsFromServer: string[] = [
+interface Good {
+  id: string;
+  value: string;
+}
+
+const goodsFromServer: Good[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -12,13 +19,113 @@ const goodsFromServer: string[] = [
...
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+].map(good => ({
+  id: uuidv4(),
+  value: good,
+}));
```
DammneD, line 25: 👍

src/App.tsx
```diff
@@ -12,13 +19,113 @@ const goodsFromServer: string[] = [
...
+                    className={classNames(
+                      'App__item-button--add',
+                      { 'App__item-button--remove': isGoodSelected },
+                    )}
+                    type="button"
+                    onClick={buttonCallback}
+                  >
+                    {buttonTitle}
+                  </button>
+                );
```
DammneD, line 108: I think you should remove this variables from `.map` at least at `render()` before `return`, but better to create a function for every logic element

DammneD (changes requested):

src/App.tsx
DammneD, line 25: 👍

solaryasha (approved):

PR #479, opened 2021-12-09
Solution

lera-permiakova:
[DEMO LINK](https://lera-permiakova.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,6 @@
+/* eslint-disable no-nested-ternary */
 import React from 'react';
+import classNames from 'classnames';
 import './App.scss';
 
 const goodsFromServer: string[] = [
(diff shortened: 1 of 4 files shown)


vlad-tuzenko (approved):
Great job  and cute design 🔥

src/App.tsx
```diff
@@ -14,11 +16,86 @@ const goodsFromServer: string[] = [
...
+            <button
+              type="button"
+              className={classNames('products__close', { active__close: selectedGood.length === 0 })}
+              disabled={selectedGood.length === 0}
+              onClick={this.clearSelectedGood}
+            >
+              X
+            </button>
+            <>
+              {this.createTitle()}
```
ihor-jpeg, line 70: you can call this function before return and assign its return value to a variable. and give it appropriate name, so that it is clear what it is

src/App.tsx
```diff
@@ -14,11 +16,86 @@ const goodsFromServer: string[] = [
...
+        <section className="products">
+          <h1 className="products__title">
+            <button
+              type="button"
+              className={classNames('products__close', { active__close: selectedGood.length === 0 })}
+              disabled={selectedGood.length === 0}
+              onClick={this.clearSelectedGood}
+            >
+              X
+            </button>
```
ihor-jpeg, line 68: this button should not be disabled, but not shown at all if no goods are selected

ihor-jpeg (changes requested):
good work! let's just fix a few things

ihor-jpeg (approved):
Well done!

PR #496, opened 2021-12-12
Solution

alinaShchavinska:
[DEMO LINK](https://alinaShchavinska.github.io/react_goods-selector/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,44 @@
-// styles go here
+ul {
+  list-style-type: none;
+}
+
+li {
+  margin-top: 4px;
+}
+
+button {
+  margin-left: 20px;
+  background-color: rgb(252, 251, 182);
+  border-radius: 3px;
+  border-color: rgb(243, 242, 163);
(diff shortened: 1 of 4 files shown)


src/App.scss
```diff
@@ -1 +1,44 @@
-// styles go here
+ul {
```
WorkAccCom, line 1: It's always good to use classes to get access to the props to avoid specificity problems.

src/App.tsx
```diff
@@ -14,11 +14,62 @@ const goodsFromServer: string[] = [
...
+            onClick={this.clearSelection}
+          >
+            X
+          </button>
+        </h1>
+        <ul className="goods__list">
+          {goodsFromServer.map((goodsName) => (
+            <li
+              key={goodsName}
+              className={this.state.goods !== goodsName ? 'good__item' : 'good__item-selected'}
```
WorkAccCom, line 57: Using classNames library would be better here

WorkAccCom (approved):

ihor-jpeg (approved):
well done!

PR #497, opened 2021-12-12
Develop

Nadiyahr:
[DEMO LINK](https://Nadiyahr.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -34,7 +34,7 @@ class App extends Component<{}, State> {
       <div className="App">
         <button
           type="button"
-          className={selected ? 'App__visible' : 'App__hiden'}
+          className={selected ? '' : 'App__hiden'}
           onClick={() => {
             this.setState({ selected: '' });
           }}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { Component } from 'react';
 import './App.scss';
 
 const goodsFromServer: string[] = [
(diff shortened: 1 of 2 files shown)


alinaShchavinska (approved):

src/App.tsx
```diff
@@ -14,11 +14,60 @@ const goodsFromServer: string[] = [
...
+  };
+
+  render() {
+    const { selected } = this.state;
+
+    return (
+      <div className="App">
+        <button
+          type="button"
+          className={selected ? '' : 'App__hiden'}
```
ArtemTopolskyi, line 37:
```suggestion
          className={selected && 'App__hiden'}
```
you can use 'classnames' library for such cases

src/App.tsx
```diff
@@ -14,11 +14,60 @@ const goodsFromServer: string[] = [
...
+          onClick={() => {
+            this.setState({ selected: '' });
+          }}
```
ArtemTopolskyi, line 40: move it to separate class method

src/App.tsx
```diff
@@ -14,11 +14,60 @@ const goodsFromServer: string[] = [
...
+            >
+              <div className="App__span">{good}</div>
+              <button
+                className="App__btn"
+                type="button"
+                onClick={() => {
+                  this.handleSelect(good);
+                }}
+              >
+                Select
```
ArtemTopolskyi, line 63: I think it will be cool to render something like 'cancel' if this good is already selected

ArtemTopolskyi (changes requested):
try to avoid anonymous function, declare it in the class body

src/App.tsx
```diff
@@ -14,11 +15,107 @@ const goodsFromServer: string[] = [
...
+              className="App__item"
+            >
+              <span className={`App__span ${selected.includes(good) && 'App__span--active'}`}>
+                {good}
+              </span>
+              <div className="App__btn">
+                <button
+                  type="button"
+                  onClick={() => {
+                    return selected.includes(good)
```
ArtemTopolskyi, line 105: why you return it?

ArtemTopolskyi (approved):

PR #503, opened 2021-12-13
Solution

aldmpr:
[DEMO LINK](https://aldmpr.github.io/react_goods-selector/)


src/App.tsx
```diff
@@ -14,11 +14,66 @@ const goodsFromServer: string[] = [
...
+  render() {
+    const { selectedGoods } = this.state;
+
+    return (
+      <div className="App">
+        <button
+          type="button"
+          onClick={() => {
+            this.clearSelect();
+          }}
```
kurtkobein, line 43:
```suggestion
          onClick={this.clearSelect}
```

kurtkobein (approved):
Great :fire: 
You missed the 12th point from the task :harold:

src/App.tsx
```diff
@@ -14,11 +14,66 @@ const goodsFromServer: string[] = [
...
 ];
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+type State = {
+  selectedGoods: string;
```
ihor-jpeg, line 18: i think this should be an array

src/App.tsx
```diff
@@ -14,11 +14,66 @@ const goodsFromServer: string[] = [
...
+          {selectedGoods
+            ? `${selectedGoods} is selected`
+            : 'No goods selected'}
+        </h1>
+
+        <ul>
+          {goodsFromServer.map(good => (
+            <li
+              key={good}
+              className={`App__item ${selectedGoods === good && 'App__item--select'}`}
```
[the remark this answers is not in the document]
ihor-jpeg: definitely

src/App.tsx
ihor-jpeg, line 43: you dont need to call the function here. just pass it and it will be called automatically on click

ihor-jpeg (changes requested):
everything works well. but i should be able to select multiple goods at the same time

ihor-jpeg (approved):

PR #526, opened 2022-01-18
first version

Niki-veb-dev:
[DEMO LINK](https://niki-veb-dev.github.io/react_goods-selector/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,53 @@
 // styles go here
+body{
+  padding: 0;
+  margin: 0;
+}
+
+.title {
+  display: flex;
+  position: absolute;
+  top: 0;
+  left: 20%;
+}
+
+.sheet {
+  margin: 0;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  &__good {
+    padding: 5px 50px;
+    text-align: center;
+  }
+}
+
+.app {
+  position: relative;
+  margin: 0 auto;
+  min-width: 768px;
+  width: 1024px;
+  display: flex;
+  justify-content: center;
+  &__list {
+    margin: 20px auto;
+    min-width: 300px;
+    padding-top: 50px;
+  }
+
+  &__item {
+    padding: 5px;
+    list-style: none;
+  }
+}
+
+.buttonVisibility {
+  visibility: hidden;
+}
+
+.button {
+  cursor: pointer;
+  padding: 5px 15px;
+  margin-right: 10px;
+  background-color: rgba(95, 245, 35, 0.705);
+}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import className from 'classnames';
 import './App.scss';
 
 const goodsFromServer: string[] = [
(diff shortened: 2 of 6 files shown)


src/App.tsx
```diff
@@ -14,11 +14,63 @@ const goodsFromServer: string[] = [
...
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
-
-export default App;
+export class App extends React.Component<{}, State> {
+  state = {
```
samson-sim, line 18:
```suggestion
  state: State = {
```

samson-sim (changes requested):

samson-sim (approved):

src/App.scss
```diff
@@ -1 +1,66 @@
...
+  display: flex;
+  min-height: 75px;
+}
+
+.sheet {
+  width: 100%;
+  margin: 0;
+  display: flex;
+  justify-content: space-between;
+  &__good {
```
Elena-brv, line 23:
```suggestion
  justify-content: space-between;
  
  &__good {
```

src/App.tsx
```diff
@@ -14,11 +14,63 @@ const goodsFromServer: string[] = [
...
-);
-
-export default App;
+export class App extends React.Component<{}, State> {
+  state: State = {
+    goods: goodsFromServer,
+    selectedGoods: ['Jam'],
+  };
+
+  getButtonWith = (text: string, setState: string[], addclassName?: string) => {
```
Elena-brv, line 23: this method is not good for understanding and there is no necessity to do it. Better create method where you need to update state and pass needed parameters onClick, but it should be 3 different buttons

src/App.tsx
```diff
@@ -14,11 +14,63 @@ const goodsFromServer: string[] = [
...
+  };
+
+  render() {
+    const { selectedGoods, goods } = this.state;
+
+    return (
+      <div className="app">
+        <h1 className="title">
+          {selectedGoods.length > 0
+            ? `${selectedGoods.join(', ')} ${selectedGoods.length > 1 ? 'are' : 'is'}     selected`
```
Elena-brv, line 44:
```suggestion
            ? `${selectedGoods.join(', ')} ${selectedGoods.length > 1 ? 'are' : 'is'} selected`
```

src/react-app-env.d.ts
```diff
@@ -1 +1,5 @@
 /// <reference types="react-scripts" />
+interface State {
```
Elena-brv, line 2: usually, state is described on the same file it is used

Elena-brv (changes requested):

src/App.tsx
Niki-veb-dev: ok

src/App.tsx
```diff
@@ -14,11 +15,98 @@ const goodsFromServer: string[] = [
...
+  };
+
+  getButtonAction = (action: string, good?: string) => {
+    if (good && action === 'select') {
+      this.setState((state) => ({ selectedGoods: [...state.selectedGoods, good] }));
+    } else if (good && action === 'remove') {
+      this.setState((state) => ({ selectedGoods: state.selectedGoods.filter(el => el !== good) }));
+    } else {
+      this.setState({ selectedGoods: [] });
+    }
```
Elena-brv, line 36:
```suggestion
    if (good) {
      switch(action) {
        case 'select':
          this.setState((state) => (
            { selectedGoods: [...state.selectedGoods, good] }
          ));
         break;
 
         case 'remove': 
           this.setState((state) => (
             { selectedGoods: state.selectedGoods.filter(el => el !== good) }
           ));
           break;
           
         default: 
           this.setState({ selectedGoods: [] });  
      }
    } else {
      this.setState({ selectedGoods: [] });
    }
```

src/App.tsx
```diff
@@ -14,11 +15,98 @@ const goodsFromServer: string[] = [
...
+        <h1 className="title">
+          {selectedGoods.length > 0
+            ? `${selectedGoods.join(', ')} ${selectedGoods.length > 1 ? 'are' : 'is'}`
+              + ' selected'
+            : 'No goods selected'}
+        </h1>
+
+        {selectedGoods.length !== 0
+          && (
+            <button
```
Elena-brv, line 53:
```suggestion
        {selectedGoods.length !== 0 && (
            <button
```

src/App.tsx
```diff
@@ -14,11 +15,98 @@ const goodsFromServer: string[] = [
...
+              type="button"
+              className="button button--big"
+              onClick={() => {
+                this.getButtonAction('ClearAll');
```
Elena-brv, line 57: do you need to put here something if method does not handle it?
better create a separate method for clearing, it may be more code but it is much easier to maintain and reuse

src/App.tsx
```diff
@@ -14,11 +15,98 @@ const goodsFromServer: string[] = [
...
+          )}
+
+        <div className="wrapper">
+          <ul className="app__list">
+            {selectedGoods.map(good => (
+              <li className="app__item" key={good}>
+                <div className="sheet">
+                  <div className="sheet__good">{good}</div>
+                  {selectedGoods.includes(good)
+                    && (
```
Elena-brv, line 71:
```suggestion
                  {selectedGoods.includes(good) && (
```

src/App.tsx
```diff
@@ -14,11 +15,98 @@ const goodsFromServer: string[] = [
...
+
+  render() {
+    const { selectedGoods, goods } = this.state;
+
+    return (
+      <div className="app">
+        <h1 className="title">
+          {selectedGoods.length > 0
+            ? `${selectedGoods.join(', ')} ${selectedGoods.length > 1 ? 'are' : 'is'}`
+              + ' selected'
```
Elena-brv, line 47: try to avoid so long lines that you should break, it's difficult to read, 
if it's impossible to place it one line, you can create a method that create a needed message

Elena-brv (changes requested):

src/App.tsx
```diff
@@ -14,11 +15,107 @@ const goodsFromServer: string[] = [
...
+            ))}
+          </ul>
+
+          <ul className="app__list">
+            {goods.map((good) => (
+              <li key={good} className="app__item">
+                <div className="sheet">
+                  <div className="sheet__good">{good}</div>
+                  {!selectedGoods.includes(good)
+                    && (
```
Elena-brv, line 102:
```suggestion
                  {!selectedGoods.includes(good) && (
```

Elena-brv (approved):

PR #555, opened 2022-01-24
goods selector solution

levix2l:
[DEMO LINK](https://levix2l.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,7 @@
 import React from 'react';
+import classNames from 'classnames';
+import 'bulma/css/bulma.min.css';
+
 import './App.scss';
 
 const goodsFromServer: string[] = [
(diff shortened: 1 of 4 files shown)


ArturErmolenko (approved):

ArturErmolenko:
🔥

src/App.tsx
```diff
@@ -14,11 +17,105 @@ const goodsFromServer: string[] = [
...
 ];
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+type State = {
+  selected: string[];
```
danheim, line 21:
```suggestion
  selectedGoods: string[];
```

src/App.tsx
```diff
@@ -14,11 +17,105 @@ const goodsFromServer: string[] = [
...
+                  <span className={classNames(
+                    {
+                      'tag is-success is-light': selected.includes(good),
+                    },
+                  )}
+                  >
+                    {good}
+                  </span>
+                  {' '}
+                  {selected.includes(good)
```
danheim, line 90: run includes method once, not twice

danheim (changes requested):

src/App.tsx
levix2l: do I need to write a new method and then run it 2 times? or where I need to save result of includes, in state?

danheim (approved):

PR #560, opened 2022-01-25
Develop

ErnestoFernandezUA:
[DEMO LINK](https://ErnestoFernandezUA.github.io/react_goods-selector/)


src/App.tsx
```diff
@@ -14,11 +15,132 @@ const goodsFromServer: string[] = [
...
+        selectedGood: [
+          ...prevState.selectedGood,
+          good,
+        ],
+      }));
+    }
+  };
+
+  removeHandler = (good: string) => {
+    const i = this.state.selectedGood.findIndex(item => item === good);
```
YevheniiKorzun, line 39: Try to use filter.

src/App.tsx
```diff
@@ -14,11 +15,132 @@ const goodsFromServer: string[] = [
...
+  displaySelectedGoods = () => {
+    if (this.state.selectedGood.length === 0) {
+      return 'No goods selected';
+    }
+
+    if (this.state.selectedGood.length === 1) {
+      return `${this.state.selectedGood[0]} is selected`;
+    }
+
+    const result = [];
```
YevheniiKorzun, line 59: There is no need to create this variable. Apply the join method to the selectedGoods array.

src/App.tsx
```diff
@@ -14,11 +15,132 @@ const goodsFromServer: string[] = [
...
+                >
+                  Add
+                </button>
+              )}
+              <span
+                className={classNames('', { selectedGood: selectedGood.includes(good) })}
+              >
+                {good}
+              </span>
+              {selectedGood.includes(good)
```
YevheniiKorzun, line 125: Try to combine checks "includes" and "not includes" into one check. You can use ternary operators.

YevheniiKorzun (changes requested):
Good job! But let's try to make it better!

YevheniiKorzun (approved):

danheim (approved):

PR #603, opened 2022-03-18
react_goods-selector

ievgen-volkov:


src/App.tsx
```diff
@@ -14,11 +15,108 @@ const goodsFromServer: string[] = [
...
+              </button>
+            )}
+          </div>
+
+          <ul className="App__list">
+            {
+              goods.map((good) => (
+                <>
+                  <li
+                    key={goods.indexOf(good)}
```
aliceblunt8, line 95: if you use fragment inside map you need to add key to fragment, not to list item

aliceblunt8 (approved):

PR #655, opened 2022-05-08
goods selector was created

Anastasiia-Svintsova:


src/App.tsx
```diff
@@ -14,11 +14,87 @@ const goodsFromServer: string[] = [
...
+    return (
+      <div className="App">
+        <div className="goods">
+          <div className="goods__selected">
+            <h1 className="goods__selected-message">{message}</h1>
+
+            <button
+              className={this.state.selectedGoods.length
+                ? 'button'
+                : 'button button--hidden'}
```
SeKaM0, line 62: You can also use `classnames` library and this code will lookes better

SeKaM0 (approved):
Good design, good job

src/App.tsx
```diff
@@ -14,11 +15,85 @@ const goodsFromServer: string[] = [
...
+                <span>{good}</span>
+
+                <button
+                  className="button"
+                  type="button"
+                  onClick={() => {
+                    this.isSelected(good);
+                  }}
+                >
+                  {this.state.selectedGoods.includes(good)
```
AlinaMatuschak, line 88: Think how you can check just one time includes

src/App.tsx
```diff
@@ -14,11 +15,85 @@ const goodsFromServer: string[] = [
...
+              type="button"
+              onClick={() => (
+                this.setState({ selectedGoods: [] })
+              )}
+            >
+              Clear
+            </button>
+          </div>
+
+          <ul className="goods__list">
```
AlinaMatuschak, line 72: you don’t need render list if goodsFromServer is empty

AlinaMatuschak (approved):

PR #659, opened 2022-05-09
first variant

serhii-naumenko:
https://serhii-naumenko.github.io/react_goods-selector


src/App.tsx
```diff
@@ -14,11 +16,111 @@ const goodsFromServer: string[] = [
...
-    {goodsFromServer.length}
-  </div>
-);
+type Props = {};
+
+type State = {
+  selectedGood: string[];
+};
+
+class App extends React.Component<Props, State> {
```
YegorKochetkov, line 25: also you could write like this:
```suggestion
type State = {
  selectedGood: string[];
};

class App extends React.Component<{}, State> {
```

src/App.tsx
```diff
@@ -14,11 +16,111 @@ const goodsFromServer: string[] = [
...
+  state: State = {
+    selectedGood: ['Jam'],
+  };
+
+  selectHandler = (good: string) => {
+    this.setState((prevState) => {
+      const newState = [...prevState.selectedGood];
+
+      if (prevState.selectedGood.includes(good)) {
+        newState.splice(newState.indexOf(good), 1);
```
YegorKochetkov, line 35: you iterate through array twice - indexOf and splice.
maybe better to use filter in this case

src/App.tsx
```diff
@@ -14,11 +16,111 @@ const goodsFromServer: string[] = [
...
+    if (selectedGood.length === 1) {
+      title = `${selectedGood} is selected`;
+    }
+
+    if (selectedGood.length >= 2) {
+      const firstPart = selectedGood.slice(0, -1).join(', ');
+      const secondPart = selectedGood.slice(-1);
+
+      title = `${firstPart} and ${secondPart} are selected`;
+    }
```
YegorKochetkov, line 67: better move this to separate method and call it in h1

src/App.tsx
```diff
@@ -14,11 +16,111 @@ const goodsFromServer: string[] = [
...
+                )}
+              >
+                {good}
+              </p>
+              <button
+                type="button"
+                className={classNames(
+                  'app__item-button',
+                  {
+                    'app__item-button--selected': selectedGood.includes(good),
```
YegorKochetkov, line 109: you twice iterate over array in render()
maybe better move it to separate `const isIncludeGood`

src/reset.scss
```diff
@@ -0,0 +1,98 @@
...
+mark,
+audio,
+video {
+  margin: 0;
+  padding: 0;
+  border: 0;
+  font: inherit;
+  font-size: 100%;
+  vertical-align: baseline;
+}
```
YegorKochetkov, line 98: why is it in such a small app?

YegorKochetkov (changes requested):
Good, but move small improvements would not hurt)

src/App.tsx
serhii-naumenko: Yes, but I decided that in my way it would be easier to change props, if necessary.

src/reset.scss
serhii-naumenko: It's just file reset. I didn't change in it anything.

Andr1yK (approved):

PR #690, opened 2022-05-31
solution of main task

EnjSadman:
[DEMO LINK](https://EnjSadman.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -14,11 +14,77 @@ const goodsFromServer: string[] = [
   'Garlic',
 ];
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+class App extends React.Component {
+  state = {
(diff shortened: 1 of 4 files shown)


vlasiuk-anatolii (approved):
Well done!

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -14,11 +14,163 @@ const goodsFromServer: string[] = [
...
+                  (this.state.selectedGoods.includes(el)
+                    ? 'goods__list-item-text-selected goods__list-item'
+                    : 'goods__list-item')
+                }
+                key={el}
+              >
+                {el}
+                <button
+                  type="button"
+                  onClick={() => {
```
Anna-Tkachenko, line 72: Will be better to create separate function in your class with this logic and add only link on this function in the jsx

Anna-Tkachenko (approved):
Good job! 👍

PR #691, opened 2022-05-31
not the final version!

vlad-demchuk:
[DEMO LINK](https://vlad-demchuk.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -24,6 +24,24 @@ class App extends React.Component<{}, State> {
     currentGoods: ['Jam'],
   };
 
+  printGoods = (goods: string[]) => {
+    if (goods.length === 0) {
+      return 'No goods selected';
+    }
+
+    if (goods.length === 1) {
+      return `${goods[0]} is selected`;
+    }
+
+    if (goods.length === 2) {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -14,11 +15,102 @@ const goodsFromServer: string[] = [
...
+              </button>
+            )}
+          </h1>
+          <h2 className="goods__list">
+            {currentGoods.length === 0 && 'No goods selected'}
+            {currentGoods.length === 1 && `${currentGoods[0]} is selected`}
+            {currentGoods.length === 2 && `${currentGoods.join(' and ')} are selected`}
+            {currentGoods.length > 2 && `${currentGoods
+              .slice(0, -1)
+              .join(', ')} and ${currentGoods[currentGoods.length - 1]} are selected`}
```
Anna-Tkachenko, line 73: Will be better to create separate function that return correct string with selected goods

Anna-Tkachenko (changes requested):
Well Done! ⭐️
Just need one small improvement)

src/App.tsx
vlad-demchuk: thanks, done)

PR #692, opened 2022-06-01
solution

ZUBOV-ILLIA:
[DEMO LINK](https://ZUBOV-ILLIA.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -26,61 +26,35 @@ class App extends React.Component<{}, State> {
   addActiveClass = 'list_item--active';
 
   render() {
+    const { selectedGood } = this.state;
+
     return (
       <>
         <div className="App">
           <h1>
-            {this.state.selectedGood
-              ? (`Selected good: - ${this.state.selectedGood}`)
+            {selectedGood
+              ? (`Selected good: - ${selectedGood}`)
               : 'No goods selected'}
           </h1>
 
-          <button
-            type="button"
-            className="clear-btn list_btn"
-            onClick={() => {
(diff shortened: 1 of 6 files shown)


vlad-demchuk:
try to implement the (* Optional) Advanced task)

src/App.tsx
```diff
@@ -14,11 +14,109 @@ const goodsFromServer: string[] = [
...
+
+          <button
+            type="button"
+            className="clear-btn list_btn"
+            onClick={() => {
+              this.setState({ selectedGood: '' });
+            }}
+          >
+            Clear
+          </button>
```
RomekIvantsiv, line 46: The clear button should not be visible if there is no selected good

src/App.tsx
```diff
@@ -14,11 +14,109 @@ const goodsFromServer: string[] = [
...
+
+  addActiveClass = 'list_item--active';
+
+  render() {
+    return (
+      <>
+        <div className="App">
+          <h1>
+            {this.state.selectedGood
+              ? (`Selected good: - ${this.state.selectedGood}`)
```
RomekIvantsiv, line 34: You can destruct to not write all the time `this.state...`, You will just use your variable in every place instead of `this.state...`

RomekIvantsiv (changes requested):
Good job,  add just small fixes 😉👍

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job! 🔥

PR #700, opened 2022-06-02
result of 2days in pain with Handling events

AndriiRut:
[DEVO LINK](https://andriirut.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -26,11 +26,9 @@ class App extends React.Component<{}, State> {
 
   buttonAction = (good: string): void => {
     if (!this.state.selectedGoods.includes(good)) {
-      this.setState((state) => {
-        state.selectedGoods.push(good);
-
-        return { selectedGoods: state.selectedGoods };
-      });
+      this.setState((state) => ({
+        selectedGoods: [...state.selectedGoods, good],
+      }));
     } else {
       this.setState((state) => {
         const newList = state.selectedGoods.filter(item => item !== good);


src/App.tsx
```diff
@@ -14,11 +15,116 @@ const goodsFromServer: string[] = [
...
+    }
+  };
+
+  messageLine = (list: string[] | []) => {
+    if (list.length === 0) {
+      return 'No goods selected';
+    }
+
+    if (list.length === 1) {
+      return `${list.join(', ')} is selected`;
```
valerii-derkach, line 49: you don't need to use `.join` here, you can find your only element from array by index `[0]`

src/App.tsx
```diff
@@ -14,11 +15,116 @@ const goodsFromServer: string[] = [
...
+              { 'is-invisible': (selectedGoods.length === 0) },
+            )}
+          >
+            Clear
+          </button>
+        </div>
+        <ul
+          className="
+          columns
+          "
```
valerii-derkach, line 91: it looks not good. can you write thr only className in one line?

src/App.tsx
```diff
@@ -14,11 +15,116 @@ const goodsFromServer: string[] = [
...
+                  className={cn(
+                    'is-medium',
+                    { 'has-background-success-light': selectedOrNot(good) },
+                  )}
+                >
+                  {good}
+                </li>
+                <button
+                  type="button"
+                  onClick={(event) => event.target && this.buttonAction(good)}
```
valerii-derkach, line 107: we shouldn't use event object in this task
just `onClick={() => this.buttonAction(good)}`

valerii-derkach (changes requested):
Good job!
but i have found some places to improve your code)

src/App.tsx
```diff
@@ -14,11 +15,116 @@ const goodsFromServer: string[] = [
...
+    const index = result.lastIndexOf(', ');
+
+    return `${result.slice(0, index)} and ${result.slice(index + 2)} are selected`;
+  };
+
+  clearButton = () => {
+    this.setState({ selectedGoods: [] });
+  };
+
+  render(): React.ReactNode {
```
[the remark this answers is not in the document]
AndriiRut: it's true, that part of code was added by VSCode automatically!

src/App.tsx
AndriiRut: My fail! In previase, there were several class for more styles but I've deleted them!

src/App.tsx
AndriiRut: I've used event object may be cause I misunderstand something in Misha's video. But now I clearly see that I was wrong. Thx for your advices!=)

Denchepornyuk (changes requested):

Anna-Tkachenko (approved):

PR #701, opened 2022-06-02
React Goods Selector

59MrRobot:
- [DEMO LINK](https://59MrRobot.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -14,11 +14,103 @@ const goodsFromServer: string[] = [
   'Garlic',
 ];
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+type State = {
+  selectedGoods: string[];
+};
+
+class App extends React.Component<{}, State> {
+  state = {
+    selectedGoods: ['Jam'],
+  };
+
+  setSelectedGoods = (good: string) => {
+    const { selectedGoods } = this.state;
+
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -14,11 +14,106 @@ const goodsFromServer: string[] = [
...
+
+      return { selectedGoods: [...prevState.selectedGoods, good] };
+    }));
+  };
+
+  clearSelectedGood = (good: string) => {
+    this.setState((prevState => {
+      return {
+        selectedGoods: [...prevState.selectedGoods].filter(e => e !== good),
+      };
```
efdalyalcin, line 45: why not implicit return? I found out today that it's a big deal :)

efdalyalcin, line 45: why not implicit return? I found out today that it's a big deal :)

src/App.tsx
```diff
@@ -14,11 +14,106 @@ const goodsFromServer: string[] = [
...
+    }));
```
efdalyalcin, line 46: why not implicit return? I found out today that it's a big deal :)

```suggestion
    this.setState((prevState => ({
        selectedGoods: [...prevState.selectedGoods].filter(e => e !== good),
      };
    )));
```

efdalyalcin (approved):
Perfect! Just one little comment. 

I also think you should look at bulma rather than writing many lines of scss. it's hard to find out what is where at first; but when you get used to it, it really makes easy styling for this kind of stuff.

meforoma (approved):
awesome 👍

PR #703, opened 2022-06-02
Solved 'React goods selector' task

dellachaise:
[DEMO LINK](https://dellachaise.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -14,11 +14,70 @@ const goodsFromServer: string[] = [
   'Garlic',
 ];
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+type State = {
+  selectedGood: string,
+};
+
+class App extends React.Component<{}, State> {
+  state = {
+    selectedGood: 'Jam',
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <h1 className="App__title">{this.state.selectedGood.length > 1 ? `${this.state.selectedGood} is selected` : 'No goods selected'}</h1>
+        <button
+          type="button"
+          className={`App__clearButton button is-light ${this.state.selectedGood === '' ? 'App__clearButton--hide is-hidden' : ''}`}
+          onClick={() => {
+            this.setState({ selectedGood: '' });
+          }}
+        >
+          Clear
+        </button>
+        <ul className="App__list">
+          {goodsFromServer.map((product) => (
+            <div key={product}>
+              <li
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -14,11 +14,70 @@ const goodsFromServer: string[] = [
...
+
+class App extends React.Component<{}, State> {
+  state = {
+    selectedGood: 'Jam',
+  };
+
+  render() {
+    return (
+      <div className="App">
+        <h1 className="App__title">{this.state.selectedGood.length > 1 ? `${this.state.selectedGood} is selected` : 'No goods selected'}</h1>
```
Anna-Tkachenko, line 29:
```suggestion
        <h1 className="App__title">
          {this.state.selectedGood
            ? `${this.state.selectedGood} is selected`
            : 'No goods selected'
          }
        </h1>
```

Anna-Tkachenko: Just make your code clear, and will be better to use destructuring at the start of the render method for getting main values from state

src/App.tsx
```diff
@@ -14,11 +14,70 @@ const goodsFromServer: string[] = [
...
+        <button
+          type="button"
+          className={`App__clearButton button is-light ${this.state.selectedGood === '' ? 'App__clearButton--hide is-hidden' : ''}`}
```
Anna-Tkachenko, line 32: Will be better to use classnames library
It's too easy to install it in your project and use

src/App.tsx
```diff
@@ -14,11 +14,70 @@ const goodsFromServer: string[] = [
...
+            <div key={product}>
+              <li
+                className={`App__list-item ${this.state.selectedGood === product ? 'App__list-item--selected' : ''}`}
+              >
+                {product}
+              </li>
+              <button
+                type="button"
+                id={product}
+                key={product + 2}
```
Anna-Tkachenko, line 50: delete key attribute from the button, it's redundant here

src/App.tsx
```diff
@@ -14,11 +14,70 @@ const goodsFromServer: string[] = [
...
+                className={`App__button button is-light ${this.state.selectedGood === product ? 'App__button--selected is-hidden' : ''}`}
+                onClick={(event) => {
+                  this.setState(
+                    {
+                      selectedGood: event.currentTarget.id,
+                    },
+                  );
+                }}
```
Anna-Tkachenko, line 58: You shouldn't use event object in this task
In simple js we don't know the name of the element, on which user clicked on
But in react we store this value in product variable and can use this variable in our function
```suggestion
                onClick={() => {
                  this.setState(
                    {
                      selectedGood: product,
                    },
                  );
                }}
```

Anna-Tkachenko (approved):
Good job! 🔥

PR #707, opened 2022-06-02
my goods selector

volodya-leskiv:
[DEMO LINK](https://vololya-leskiv.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -14,11 +14,84 @@ const goodsFromServer: string[] = [
   'Garlic',
 ];
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+type State = {
+  selectedGoods: string[];
+};
 
-export default App;
+export class App extends React.Component<{}, State> {
+  state: State = {
+    selectedGoods: ['Jam'],
+  };
+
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -14,11 +14,84 @@ const goodsFromServer: string[] = [
...
+      selectedGoods: [],
+    });
+  };
+
+  listToShow = () => {
+    if (this.state.selectedGoods.length === 1) {
+      return `${this.state.selectedGoods.join(', ')} is selected`;
+    }
+
+    return `${this.state.selectedGoods.join(', ')} are selected`;
```
valerii-derkach, line 49: I believe you can do it better, and improve the` listToShow` function to print `good1, good2 and lastgood` , not only a list separated with coma

valerii-derkach (approved):
Great job!
you have written so many styles manualy!
try to use css library next time
you can also improve you code, if you want)

src/App.tsx
volodya-leskiv: Thanks for the feedback, Valerii! At the expense of styles, I wanted to practice, because I worked with css for a long time) In the future I need to try some framework). At the expense of the comment, your opinion is good. Maybe on the weekend I'll come up with something))

Anna-Tkachenko (approved):
Good job! 🔥

PR #714, opened 2022-06-03
finished task

iwanttosmile4u:
[DEMO LINK](https://iwanttosmile4u.github.io/react_goods-selector/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,81 @@
-// styles go here
+body {
+  margin: 0;
+  padding: 0;
+
+  background-color: burlywood;
+}
+
+ul {
+  margin: 0;
+  padding: 0;
+}
+
+.app {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+}
+
+.title {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+
+  margin-bottom: 20px;
+
+  &__name {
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -14,11 +18,92 @@ const goodsFromServer: string[] = [
...
+
+        <ul className="goods-list">
+          {goodsFromServer.map((good) => (
+            <li className="goods-list__item" key={good}>
+              {good}
+              <button
+                className="goods-list__btn"
+                type="button"
+                onClick={() => {
+                  // eslint-disable-next-line @typescript-eslint/no-unused-expressions
```
EnjSadman, line 91: this is not right. Mb move this to other function

EnjSadman (approved):

src/App.tsx
iwanttosmile4u: I'll use if...else instead of ternary operator. Bacause I didn't understand why linter show me this proplem when I used ternary operator, so i desided to use quick fix)

src/App.scss
```diff
@@ -1 +1,81 @@
-// styles go here
+body {
+  margin: 0;
+  padding: 0;
```
Denchepornyuk, line 3: body doesn't have default padding, you can remove it

Denchepornyuk (approved):

PR #724, opened 2022-06-07
Solution

Danylo-Dziuban:
[DEMO LINK](https://Danylo-Dziuban.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable react/jsx-one-expression-per-line */
 import React from 'react';
 import './App.scss';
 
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -14,11 +15,98 @@ const goodsFromServer: string[] = [
...
+            ))}
+          </ul>
+
+          <button
+            type="button"
+            className="btn btn-danger"
+            onClick={() => this.setState({ addedGoods: [] })}
+          >
+            Clear
+          </button>
```
EnjSadman, line 96: This button should be invisible when no goods selected

src/App.tsx
```diff
@@ -14,11 +15,98 @@ const goodsFromServer: string[] = [
...
+        </div>
+
+        <div className="w-50">
+          <div className="fs-4 mb-3">Selected goods: {array}</div>
```
EnjSadman, line 100: This one should be "no goods selected" when none of goods is selected

EnjSadman (changes requested):

src/App.tsx
```diff
@@ -14,11 +15,98 @@ const goodsFromServer: string[] = [
...
+    const array = this.state.addedGoods.join(', ');
+
+    return (
+      <div className="App p-4 d-flex">
+        <div className="w-50">
+          <ul className="list-group mb-3">
+            {goodsFromServer.map(good => (
+              <li
+                key={good}
+                className={this.state.addedGoods.includes(good)
```
Anna-Tkachenko, line 37: Will be better to store the result of includes method in some variable
Now you call the same function at least 3 times in your code

Anna-Tkachenko (approved):

PR #726, opened 2022-06-08
Solution

iLebo-stack:
[DEMO LINK](https://iLebo-stack.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -14,11 +14,91 @@ const goodsFromServer: string[] = [
   'Garlic',
 ];
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+interface State {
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -14,11 +14,91 @@ const goodsFromServer: string[] = [
...
+        <h1>
+          {goodItem && h1Content}
+
+          <br />
+
+          <button
+            type="button"
+            onClick={() => {
+              this.setState({
+                goodItem: '',
```
efdalyalcin, line 48: I think you should have turn to 'No goods selected' for goodItem. Clear removes h1, but in the readme it says `clear the selection`.

efdalyalcin (changes requested):
Nice work, almost done. Please look at the comment. I also think you could have made a better looking page with simple bulma classes, but not necessary for the time being.

ihor-jpeg (approved):

PR #727, opened 2022-06-08
Goods-selector stared

Sardorbek1122:
[DEMO LINK](https://Sardorbek1122.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -24,9 +24,9 @@ class App extends React.Component<{}, State> {
   };
 
   setSelectedGoods = (goods: string) => {
-    this.setState(((prevstate: State) => {
-      return { selectedGoods: [...prevstate.selectedGoods, goods] };
-    }));
+    this.setState(((prevState: State) => (
+      { selectedGoods: [...prevState.selectedGoods, goods] }
+    )));
   };
 
   removeSelectedGood = (good: string) => {
@@ -78,7 +78,14 @@ class App extends React.Component<{}, State> {
         <ul className="select__items">
           {goodsFromServer.map(good => {
             const isSelected = selectedGoods.includes(good);
-            const buttonText = isSelected ? 'Remove' : 'Select';
+            const buttonText = isSelected
+              ? 'Remove'
+              : 'Select';
+            const handleClick = () => (
+              this.state.selectedGoods.includes(good)
+                ? this.removeSelectedGood(good)
+                : this.setSelectedGoods(good)
+            );
 
             return (
               <li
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -14,11 +14,95 @@ const goodsFromServer: string[] = [
...
+          <h1 className="select__content">
+            {this.createTitle(selectedGoods)}
+          </h1>
+
+          <button
+            type="button"
+            className="select__clearBtn"
+          >
+            clear
+          </button>
```
rHashTag, line 74: `onClick` has not been set for the button. Clear button therefore does not work.

rHashTag (changes requested):
Good job!

One comment to review.

src/App.tsx
Sardorbek1122: fixed

src/App.tsx
```diff
@@ -14,11 +14,96 @@ const goodsFromServer: string[] = [
...
+
+class App extends React.Component<{}, State> {
+  state = {
+    selectedGoods: ['Jam'],
+  };
+
+  setSelectedGoods = (goods: string) => {
+    this.setState(((prevstate: State) => {
+      return { selectedGoods: [...prevstate.selectedGoods, goods] };
+    }));
```
ihor-jpeg, line 29: why not return immediately? why do you need the function body?

src/App.tsx
```diff
@@ -14,11 +14,96 @@ const goodsFromServer: string[] = [
...
+  };
+
+  removeSelectedGood = (good: string) => {
+    this.setState((prevState => ({
+      selectedGoods: prevState.selectedGoods.filter(remaining => (
```
ihor-jpeg, line 34: be consistent, in the first one you specify `prevstate: State`, but in the second one simply `prevstate`. decide which one you prefer and be consistent

src/App.tsx
```diff
@@ -14,11 +14,96 @@ const goodsFromServer: string[] = [
...
+              >
+                <span>{good}</span>
+                <button
+                  type="button"
+                  className="select__btn"
+                  onClick={() => {
+                    return isSelected
+                      ? this.removeSelectedGood(good)
+                      : this.setSelectedGoods(good);
+                  }}
```
ihor-jpeg, line 96: create a handler function
and why do you need the body and the return keyword?

ihor-jpeg (changes requested):

meforoma (approved):

PR #730, opened 2022-06-08
process

Uktam19980416:


Uktam19980416:
[DEMO LINK](https://Uktam19980416.github.io/react_goods-selector/)

LukaszGryziewicz (approved):
Looks good 👍

src/App.tsx
```diff
@@ -14,11 +18,83 @@ const goodsFromServer: string[] = [
...
-);
+class App extends React.Component<{}, State> {
+  state = {
+    selectedGoods: ['Jam'],
+  };
+
+  selectGoods = (good: string) => {
+    this.setState(((prevState: State) => {
+      return { selectedGoods: [...prevState.selectedGoods, good] };
+    }));
```
meforoma, line 29: get used to implicit return
```suggestion
    this.setState(((prevState: State) => (
      { selectedGoods: [...prevState.selectedGoods, good] };
    )));
```

src/App.tsx
```diff
@@ -14,11 +18,83 @@ const goodsFromServer: string[] = [
...
+  };
+
+  removeGoods = (good: string) => {
+    this.setState((prevState => ({
```
meforoma, line 33: let's be consistent with type definitions - for select goods you specify `prevState` type, here - you don't.
Better stick to certain convention - either define types or not, but keep this approach across your code.
```suggestion
    this.setState((prevState: State => ({
```

src/App.tsx
```diff
@@ -14,11 +18,83 @@ const goodsFromServer: string[] = [
...
+              <li
+                key={good}
+              >
+                <span>{good}</span>
+                <button
+                  type="button"
+                  onClick={() => {
+                    return isSelected
+                      ? this.removeGoods(good)
+                      : this.selectGoods(good);
```
meforoma, line 86: implicit return
and you could introduce a handler function instead ternary-ing, so you pass only the handler
```suggestion
                  onClick={() => (
                    isSelected
                      ? this.removeGoods(good)
                      : this.selectGoods(good);
```

or
```suggestion
                  onClick={handler}
```

meforoma (changes requested):
please address comments

Uktam19980416:
[DEMO LINK](https://Uktam19980416.github.io/react_goods-selector/)

meforoma (approved):

PR #731, opened 2022-06-08
task completed

ansu16281:
[DEMO LINK](https://andreisucman.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -61,7 +61,7 @@ class App extends React.Component<{}, State> {
     return (
       <div className="App">
         <div>
-          <h1>
+          <h1 className="selectedGood">
             {`Selected good: ${this.createMessage()}`}
           </h1>
 
@@ -87,7 +87,7 @@ class App extends React.Component<{}, State> {
             })}
           </ul>
 
-          {!!this.state.items.length && (
+          {this.state.items.length && (
             <button
               type="button"
               onClick={() => {


src/App.tsx
```diff
@@ -14,11 +14,93 @@ const goodsFromServer: string[] = [
...
+                    onClick={() => this.buttonSelection(isSelected, good)}
+                  >
+                    {buttonText}
+                  </button>
+                </li>
+              );
+            })}
+          </ul>
+
+          {!!this.state.items.length && (
```
iLebo-stack, line 90: I don't really understand what the `!!` does here.

src/App.scss
```diff
@@ -1 +1,3 @@
-// styles go here
+.selectedGood {
+  background-color: lightgray;
+}
```
iLebo-stack, line 3: You have forgotten to use this class.

iLebo-stack (changes requested):

src/App.tsx
andreisucman: I think it's just a trick from OP to catch if I'm copying his code. He got me...

andreisucman: I think it's just a trick from the OP to check if I'm copying his code. He got me...

src/App.tsx
```diff
@@ -14,11 +14,93 @@ const goodsFromServer: string[] = [
...
+              );
+            })}
+          </ul>
+
+          {this.state.items.length && (
+            <button
+              type="button"
+              onClick={() => {
+                this.setState({ items: [] });
+              }}
```
ihor-jpeg, line 95: maybe a handler outside of render method

ihor-jpeg (approved):

PR #744, opened 2022-06-27
Goods selector

pavlenkoam:
[Demo](https://pavlenkoam.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -14,11 +14,84 @@ const goodsFromServer: string[] = [
   'Garlic',
 ];
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Selected good: -</h1>
-    {goodsFromServer.length}
-  </div>
-);
+type State = {
+  selectedItem: string[],
+};
+
+class App extends React.Component<{}, State> {
+  state = {
+    selectedItem: ['Jam'],
+  };
+
+  changeGoodsList = (item: string) => {
+    const { selectedItem } = this.state;
+
+    if (selectedItem.includes(item)) {
+      selectedItem.splice(selectedItem.indexOf(item), 1);
+    } else {
+      selectedItem.push(item);
+    }
+
+    this.setState({ selectedItem });
+  };
+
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -14,11 +14,84 @@ const goodsFromServer: string[] = [
...
-    {goodsFromServer.length}
-  </div>
-);
+type State = {
+  selectedItem: string[],
+};
+
+class App extends React.Component<{}, State> {
+  state = {
+    selectedItem: ['Jam'],
```
aliceblunt8, line 23:
```suggestion
    selectedItems: ['Jam'],
```

src/App.tsx
```diff
@@ -14,11 +14,84 @@ const goodsFromServer: string[] = [
...
+  };
+
+  changeGoodsList = (item: string) => {
+    const { selectedItem } = this.state;
+
+    if (selectedItem.includes(item)) {
+      selectedItem.splice(selectedItem.indexOf(item), 1);
+    } else {
+      selectedItem.push(item);
```
aliceblunt8, line 32: we don't mutate the state, we add a new one inside this.setState

aliceblunt8: https://reactjs.org/docs/react-component.html#setstate

aliceblunt8: you can use functional setState in this case

aliceblunt8 (changes requested):

src/App.tsx
```diff
@@ -14,11 +14,82 @@ const goodsFromServer: string[] = [
...
+
+  render() {
+    const { selectedItems } = this.state;
+
+    return (
+      <div className="App">
+        <h1>
+          {this.setHeader()}
+          <button
+            className={`clearButton ${selectedItems.length === 0 ? 'hidden' : ''}`}
```
alexandra-protyanova, line 62: u can try to use classnames library

alexandra-protyanova (approved):

PR #748, opened 2022-07-04
Solution

mira1zu:
- [DEMO LINK](https://mira1zu.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,30 @@
 import React from 'react';
+
 import './App.scss';
+
+function makeStringOfProducts(array: string[]): string {
+  if (array.length === 0) {
+    return 'No goods selected';
+  }
+
+  if (array.length === 1) {
+    return `${array[0]} is selected`;
+  }
+
+  let string = '';
+
+  for (let i = 0; i < array.length - 1; i += 1) {
+    if (i > 0) {
+      string += ', ';
+    }
+
+    string += array[i];
+  }
+
+  string += ` and  ${array[array.length - 1]}`;
+
+  return `${string} are selected`;
+}
 
 const goodsFromServer: string[] = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,6 +1,31 @@
...
+    if (i > 0) {
+      string += ', ';
+    }
+
+    string += array[i];
+  }
+
+  string += ` and  ${array[array.length - 1]}`;
+
+  return `${string} are selected`;
```
jstmpelowycz, line 26: I guess you could avoid doing so
hint: try using a couple of `slice` + `join`

src/App.tsx
```diff
@@ -14,11 +39,99 @@ const goodsFromServer: string[] = [
...
+      const index = selectedProducts.indexOf(product);
+      const newProducts = selectedProducts
+        .slice(0, index)
+        .concat(selectedProducts.slice(index + 1));
+
+      return ({
+        selectedProducts: [...newProducts],
+      });
+    });
+  };
```
jstmpelowycz, line 73: perhaps, filter is a better solution?

src/App.tsx
```diff
@@ -14,11 +39,99 @@ const goodsFromServer: string[] = [
...
+    const { listOfProducts, selectedProducts } = this.state;
+
+    return (
+      <div className="App">
+        <div className="list container box">
+          <ul>
+            {listOfProducts.map((product) => (
+              <li
+                key={product}
+                className={`list__item ${selectedProducts.includes(product) ? 'list__item--selected' : ''}`}
```
jstmpelowycz, line 89: giving empty string a ternary result is not the best solution, for sure
you could try this one, for example: https://www.npmjs.com/package/classnames
it is frequently used in real projects

src/App.tsx
```diff
@@ -14,11 +39,99 @@ const goodsFromServer: string[] = [
...
+                <p className="product is-size-2">
+                  {product}
+                </p>
+
+                {selectedProducts.includes(product)
+                  ? (
+                    <button
+                      className="button is-primary is-medium"
+                      type="button"
+                      onClick={() => this.removeProduct(product)}
```
jstmpelowycz, line 100:
```suggestion
                      onClick={this.removeProduct}
```

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #817, opened 2022-08-01
Solution

fledith:
[DEMO LINK](https://fledith.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,51 +1,127 @@
 import React from 'react';
+import classNames from 'classnames';
+import { v4 as uuidv4 } from 'uuid';
 import './App.scss';
 
-// import goodsFromServer from './goods';
+import goodsFromServer from './goods';
 
-export const App: React.FC = () => {
-  return (
-    <main className="App">
-      <header className="App__header">
-        <h1 className="App__title">
-          Jam is selected
-        </h1>
+const uniqueKeys: string[] = [];
 
-        <button
-          type="button"
-          className="App__clear"
-        >
-          Clear
-        </button>
-      </header>
+for (let i = 0; i < goodsFromServer.length; i += 1) {
+  uniqueKeys[i] = uuidv4();
+}
 
-      <ul>
-        <li className="Good">Dumplings</li>
-        <li className="Good">Carrot</li>
-        <li className="Good">Eggs</li>
-        <li className="Good">Ice cream</li>
-        <li className="Good">Apple</li>
-        <li className="Good">Bread</li>
-        <li className="Good">Fish</li>
-        <li className="Good">Honey</li>
-        <li className="Good Good--active">Jam</li>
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,51 +1,127 @@
...
-      </button>
-    </main>
-  );
+import goodsFromServer from './goods';
+
+const uniqueKeys: string[] = [];
+
+for (let i = 0; i < goodsFromServer.length; i += 1) {
+  uniqueKeys[i] = uuidv4();
+}
```
BudnikOleksii, line 12: You can add uniqueKeys to the current array of goods

src/App.tsx
```diff
@@ -1,51 +1,127 @@
...
+
+    if (length > 1) {
+      this.setState({ title: `${selectedGoods.slice(0, length - 1).join(', ')} and ${selectedGoods[length - 1]} are selected` });
+    }
+  };
+
+  handlingRemove = (good: string, goods: string[]) => {
+    this.setState(
+      { selectedGoods: goods.filter(item => item !== good) },
+      this.setTitle,
```
BudnikOleksii, line 49: All handlers call the setTitle method, maybe it's better to call it inside render, it will work in all cases

BudnikOleksii (changes requested):
Let's improve a little bit

src/App.tsx
fledith: I think it is not necessary for this task

src/App.tsx
fledith: not agree

solaryasha (changes requested):

src/App.tsx
```diff
@@ -1,51 +1,127 @@
...
+
+type State = {
+  selectedGoods: string[];
+  title: string;
 };
+
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    selectedGoods: ['Jam'],
+    title: '',
```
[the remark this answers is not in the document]
fledith: bcz it says so in the task

src/App.tsx
```diff
@@ -1,51 +1,127 @@
...
+  };
+
+  componentDidMount() {
+    this.setTitle();
+  }
```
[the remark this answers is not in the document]
solaryasha: This comment still not fixed

solaryasha (changes requested):

src/App.tsx
```diff
@@ -1,51 +1,118 @@
...
+            ? (
+              <button
+                type="button"
+                className="App__clear button is-danger"
+                onClick={this.handlingClear}
+              >
+                Clear
+              </button>
+            )
+            : '' }
```
solaryasha, line 73:
```suggestion
            : null}
```

solaryasha (approved):

PR #822, opened 2022-08-02
solution

oleksandr-kotliarov:
[DEMO LINK](https://oleksandr-kotliarov.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,51 +1,99 @@
 import React from 'react';
 import './App.scss';
+import classNames from 'classnames';
 
-// import goodsFromServer from './goods';
+import goodsFromServer from './goods';
 
-export const App: React.FC = () => {
-  return (
-    <main className="App">
-      <header className="App__header">
-        <h1 className="App__title">
-          Jam is selected
-        </h1>
-
-        <button
-          type="button"
-          className="App__clear"
-        >
-          Clear
-        </button>
-      </header>
-
-      <ul>
-        <li className="Good">Dumplings</li>
-        <li className="Good">Carrot</li>
-        <li className="Good">Eggs</li>
-        <li className="Good">Ice cream</li>
-        <li className="Good">Apple</li>
-        <li className="Good">Bread</li>
-        <li className="Good">Fish</li>
-        <li className="Good">Honey</li>
-        <li className="Good Good--active">Jam</li>
-        <li className="Good">Garlic</li>
-      </ul>
-
-      {/* Put required buttons into each Good */}
-      <button
-        type="button"
-        className="Good__remove"
-      >
-        Remove
-      </button>
-
-      <button
-        type="button"
-        className="Good__select"
-      >
-        Select
-      </button>
-    </main>
-  );
+type State = {
+  selectedGoods: string[];
 };
+
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    selectedGoods: ['Jam'],
+  };
+
+  removeEl = (list: string[], good: string) => {
+    return list.filter((el: string) => el !== good);
+  };
+
+  titleDescription = (list: string[]) => {
+    switch (list.length) {
+      case 0:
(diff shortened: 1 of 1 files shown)


VETALEK (changes requested):
GJ 👍 
need to fix this
image
image

src/App.tsx
```diff
@@ -1,51 +1,103 @@
...
+
+          {selectedGoods.length
+            ? (
+              <button
+                type="button"
+                className="App__clear"
+                onClick={() => {
+                  this.setState({ selectedGoods: [] });
+                }}
+              >
```
solaryasha, line 51: Let's create separate method for this, so we don't create new arrow function after component re-rendering

src/App.tsx
```diff
@@ -1,51 +1,103 @@
...
+                  )
+                  : (
+                    <button
+                      type="button"
+                      className="Good__select"
+                      onClick={() => {
+                        this.setState(
+                          { selectedGoods: [...selectedGoods, good] },
+                        );
+                      }}
```
solaryasha, line 91: Here as well

src/App.tsx
```diff
@@ -1,51 +1,103 @@
...
+            return (
+              <li
+                key={good}
+                className={classNames('Good', { 'Good--active': isActive })}
+              >
+                {good}
+
+                {isActive
+                  ? (
+                    <button
```
solaryasha, line 71: Let's create separate Button component that takes:

1. text
2. onClick callback

src/App.tsx
```diff
@@ -1,51 +1,103 @@
...
+type State = {
+  selectedGoods: string[];
 };
+
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    selectedGoods: ['Jam'],
+  };
+
+  removeEl = (list: string[], good: string) => {
```
solaryasha, line 16: Don't shorten method's name

solaryasha: Let's create method with `setState`

src/App.tsx
```diff
@@ -1,51 +1,103 @@
...
+    return list.filter((el: string) => el !== good);
+  };
+
+  titleDescription = (list: string[]) => {
```
solaryasha, line 20: Method's name should start with a verb

solaryasha (changes requested):

src/App.tsx
oleksandr-kotliarov: image
?

solaryasha: You are right, forget about this condition :)

src/App.tsx
```diff
@@ -1,51 +1,113 @@
...
+type State = {
+  selectedGoods: string[];
 };
+
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    selectedGoods: ['Jam'],
+  };
+
+  addSelected = (good: string) => {
```
solaryasha, line 16:
```suggestion
  addSelectedGood = (good: string) => {
```

src/App.tsx
```diff
@@ -1,51 +1,113 @@
...
+    const { selectedGoods } = this.state;
+
+    this.setState(
+      { selectedGoods: [...selectedGoods, good] },
+    );
+  };
+
+  removeSelected = (good: string) => {
```
solaryasha, line 24:
```suggestion
  removeSelectedGood = (good: string) => {
```

src/App.tsx
```diff
@@ -1,51 +1,113 @@
...
+          <h1 className="App__title">
+            {this.pickTitleDescription()}
+          </h1>
+
+          {selectedGoods.length
+            ? (
+              <button
+                type="button"
+                className="App__clear"
+                onClick={() => this.clearSelectedList()}
```
solaryasha, line 68:
```suggestion
                onClick={this.clearSelectedList}
```

solaryasha (approved):

PR #825, opened 2022-08-04
Solution

TwoAndMore:
[DEMO LINK](https://TwoAndMore.github.io/react_goods-selector/)


fledith (changes requested):
image
u have to hide the button clear when no one good is selected

PR #839, opened 2022-08-08
solution

ClaudiaKalnitskaya:
[DEMO LINK](https://ClaudiaKalnitskaya.github.io/react_goods-selector/)


illnino380 (approved):
I like your solution but didn't get why you use switch/case instead of if checks)

ivan-kadykalo (approved):
GJ🔥

PR #882, opened 2022-09-03
Solution

11Olya112:
https://github.com/mate-academy/react_goods-selector/compare/master...11Olya112:react_goods-selector:develop

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,51 +1,84 @@
 import React from 'react';
 import './App.scss';
+import classnames from 'classnames';
+import goodsFromServer from './goods';
 
-// import goodsFromServer from './goods';
-
-export const App: React.FC = () => {
-  return (
-    <main className="App">
-      <header className="App__header">
-        <h1 className="App__title">
-          Jam is selected
-        </h1>
-
-        <button
-          type="button"
-          className="App__clear"
-        >
-          Clear
-        </button>
-      </header>
-
-      <ul>
-        <li className="Good">Dumplings</li>
-        <li className="Good">Carrot</li>
-        <li className="Good">Eggs</li>
-        <li className="Good">Ice cream</li>
-        <li className="Good">Apple</li>
-        <li className="Good">Bread</li>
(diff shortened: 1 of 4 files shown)


SanyaBratashchuk (changes requested):
Add empty state
image

Bohdan-Maksymiuk (changes requested):
https://user-images.githubusercontent.com/94289038/188563973-54e43785-db9f-4309-a935-5763c458aa92.mov

Bohdan-Maksymiuk (changes requested):
Screenshot 2022-09-07 at 09 10 50

the phrase should be `No goods selected`
you don't need to store it in state `selectedGood`, because this is not good
store empty string or null here, and if this state is falsy - show our string instead of selected good

src/App.tsx
```diff
@@ -1,51 +1,84 @@
...
-        <li className="Good">Eggs</li>
-        <li className="Good">Ice cream</li>
-        <li className="Good">Apple</li>
-        <li className="Good">Bread</li>
-        <li className="Good">Fish</li>
-        <li className="Good">Honey</li>
-        <li className="Good Good--active">Jam</li>
-        <li className="Good">Garlic</li>
-      </ul>
+    const selgoods = selectedGood ? `${selectedGood} ${' '} is selected` : 'No goods selected';
```
Bohdan-Maksymiuk, line 18:
```suggestion
    const heading = selectedGood
      ? `${selectedGood} ${' '} is selected`
      : 'No goods selected';
```
heading name better suits here
and better to format code like this - it is easier to read

Bohdan-Maksymiuk, line 18:
```suggestion
    const selgoods = selectedGood ? `${selectedGood} is selected` : 'No goods selected';
```

Bohdan-Maksymiuk: and you don't need this `${' '} ` in template

Bohdan-Maksymiuk (approved):
Good job 👍

PR #887, opened 2022-09-06
Solution

Dmytro-Matsehora:
[DEMO LINK](https://Dmytro-Matsehora.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,94 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string,
+  className: string,
+  blueLight: string,
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    selectedGood: 'Jam',
+    className: 'has-background-success-light',
+    blueLight: 'is-info',
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  render() {
+    const {
+      selectedGood,
+      className,
+      blueLight,
+    } = this.state;
 
-    <table className="table">
-      <tbody>
-        <tr data-cy="Good">
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+    return (
+      <main className="section container">
+        {selectedGood === ''
+          ? <h1 className="title">No goods selected</h1>
+          : (
+            <h1 className="title is-flex is-align-items-center">
+              {`${selectedGood} is selected`}
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+              <button
+                data-cy="ClearButton"
+                type="button"
+                className="delete ml-3"
+                onClick={() => this.setState(
+                  { selectedGood: '' },
+                )}
+              />
+            </h1>
+          )}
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +15,94 @@ export const goods = [
...
+  blueLight: string,
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    selectedGood: 'Jam',
+    className: 'has-background-success-light',
+    blueLight: 'is-info',
```
AlexanderKolomiiets, line 28: in my opinion, it is unnecessary to set classes in state

src/App.tsx
```diff
@@ -15,71 +15,94 @@ export const goods = [
...
-          </td>
-        </tr>
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+              <button
+                data-cy="ClearButton"
+                type="button"
+                className="delete ml-3"
+                onClick={() => this.setState(
+                  { selectedGood: '' },
+                )}
```
AlexanderKolomiiets, line 53: you gonna use the same handler in your code later, so it is better to put a handler in a method

src/App.tsx
```diff
@@ -15,71 +15,94 @@ export const goods = [
...
-              data-cy="RemoveButton"
-              type="button"
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
```
AlexanderKolomiiets, line 60: checklist point 4

src/App.tsx
```diff
@@ -15,71 +15,94 @@ export const goods = [
...
+                          : ''
+                      }`
+                    }
+                    onClick={() => (
+                      selectedGood === good
+                        ? this.setState({ selectedGood: '' })
+                        : this.setState({
+                          selectedGood: good,
+                          className: 'has-background-success-light',
+                          blueLight: 'is-info',
```
AlexanderKolomiiets, line 90: in this case, you don't have to write it because in your code you have never changed `className` and `blueLight` in your state

AlexanderKolomiiets (changes requested):
Great, but add some fixes

dellachaise (commented):
Check your task according to the checklist and put the mark on the table.
Screenshot 2022-09-13 at 12 44 37

AlexanderKolomiiets (approved):

src/App.tsx
```diff
@@ -15,71 +15,92 @@ export const goods = [
...
+          ? <h1 className="title">No goods selected</h1>
+          : (
+            <h1 className="title is-flex is-align-items-center">
+              {`${selectedGood} is selected`}
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Jam
-          </td>
-        </tr>
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
dellachaise, line 53: Try to rewrite without `eslint-disable`.

src/App.tsx
```diff
@@ -15,71 +15,92 @@ export const goods = [
...
+                        ? 'RemoveButton'
+                        : 'AddButton'
+                    }
+                    type="button"
+                    className={
+                      `button ${
+                        selectedGood === good
+                          ? blueLight
+                          : ''
+                      }`
```
dellachaise, line 88: Better install [classnames](https://www.npmjs.com/package/classnames) and use it instead of the ternary operator.
The ternary operator is very limited in its functionality to be used in classNames. 
[classnames](https://www.npmjs.com/package/classnames) usually used in real projects and would be great to work with it now and would be needed to do in test tasks.

src/App.tsx
```diff
@@ -15,71 +15,92 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string,
+  blueLight: string,
```
dellachaise, line 20: Why did you same it in the state? Try to rewrite without storing this value on the state.

dellachaise (approved):
Great job 👍 

PR #893, opened 2022-09-10
Solution

VolodymyrChuyko:
[DEMO LINK](https://VolodymyrChuyko.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


andrii-dmytruk (approved):
Well done!

src/App.tsx
```diff
@@ -15,71 +16,92 @@ export const goods = [
...
-              -
-            </button>
-          </td>
+              className="delete ml-3"
+              onClick={this.clearSelected}
+            />
+          </h1>
+        ) : (
+          <h1 className="title">No goods selected</h1>
+        )}
```
denys-danyliuk, line 55: only content inside h1 tag changed depending if there is selectedGood or not, so it makes no sense when you render the same component, just change content depending on selectedGood

src/App.tsx
```diff
@@ -15,71 +16,92 @@ export const goods = [
...
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Jam
-          </td>
-        </tr>
+        <table className="table">
+          <tbody>
+            {goods.map(good => {
+              const isSelected = selectedGood === good;
```
denys-danyliuk, line 60: do you really need a variable for this?

denys-danyliuk (changes requested):
Nicely done! Pay attention to the comments

src/App.tsx
VolodymyrChuyko: I use this condition twice in my code. So I decide to make a variable for better maintenance and readability. However, I get rid of the 'selectedBackground' variable  for className

denys-danyliuk (approved):
🔥

PR #899, opened 2022-09-12
Solution

Anastasiia-Khmelovska:
[DEMO LINK](https://Anastasiia-Khmelovska.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,84 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string;
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component {
+  state: Readonly<State> = {
+    selectedGood: 'Jam',
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  render() {
+    const { selectedGood } = this.state;
 
-    <table className="table">
-      <tbody>
-        <tr data-cy="Good">
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+    return (
+      <main className="section container">
+        {selectedGood.length === 0
+          ? <h1 className="title">No goods selected</h1>
+          : (
+            <h1 className="title is-flex is-align-items-center">
+              {selectedGood}
+              {' '}
+              is selected
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+              <button
+                data-cy="ClearButton"
+                type="button"
+                className="delete ml-3"
+                onClick={() => {
+                  this.setState({ selectedGood: '' });
+                }}
+              />
+            </h1>
+          )}
 
-        <tr data-cy="Good" className="has-background-success-light">
-          <td>
-            <button
-              data-cy="RemoveButton"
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +15,84 @@ export const goods = [
...
+          <tbody>
+            {goods.map(good => (
+              <tr
+                data-cy="Good"
+                className={selectedGood !== good
+                  ? ''
+                  : 'has-background-success-light'}
+              >
+                <td>
+                  {selectedGood === good
```
DimaKobzar7, line 64: Use here classNames library instead of ternary operator
` className={selectedGood !== good
                  ? ''
                  : 'has-background-success-light'}
              >
                <td>`

DimaKobzar7 (approved):

src/App.tsx
dellachaise: Agreed.
The ternary operator is very limited in its functionality to be used in classNames. (this way usually did in real projects and would be great to work with it now and would be needed to do in test tasks).

src/App.tsx
```diff
@@ -15,71 +15,84 @@ export const goods = [
...
-          <td>
-            <button
-              data-cy="RemoveButton"
-              type="button"
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
dellachaise, line 44: Do not disable es-lint.
Why is your button inside `h1` tag?

src/App.tsx
```diff
@@ -15,71 +15,84 @@ export const goods = [
...
-  </main>
-);
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
+                data-cy="Good"
+                className={selectedGood !== good
+                  ? ''
+                  : 'has-background-success-light'}
```
dellachaise, line 61: Better install [classnames](https://www.npmjs.com/package/classnames) and use it instead of ternary operator.

src/App.tsx
```diff
@@ -15,71 +15,84 @@ export const goods = [
...
+                      </button>
+                    )
+                    : (
+                      <button
+                        data-cy="AddButton"
+                        type="button"
+                        className="button"
+                        onClick={() => {
+                          this.setState({ selectedGood: good });
+                        }}
```
dellachaise, line 82: [checklist point 2](https://github.com/mate-academy/react_goods-selector/blob/master/checklist.md)

dellachaise (changes requested):
Great start 👍  

src/App.tsx
Anastasiia-Khmelovska: It was originally in the assignment, linter does not allow without it

src/App.tsx
Anastasiia-Khmelovska: done)

src/App.tsx
dellachaise: Yes, it was originally, but without it code would be cleaner. 

Try to do one check for `h1` and the separate check for `button`.
```suggestion
              {selectedGood && (
            <button >
              -
            </button>
          )}
```

src/App.tsx
```diff
@@ -15,71 +16,84 @@ export const goods = [
...
-              data-cy="RemoveButton"
-              type="button"
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+    return (
+      <main className="section container">
+        {selectedGood.length === 0
```
dellachaise, line 41:
```suggestion
        {!selectedGood.length
```

src/App.tsx
```diff
@@ -15,71 +16,84 @@ export const goods = [
...
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  noSelectedGood = () => {
```
dellachaise, line 28: [checklist point 1](https://github.com/mate-academy/react_goods-selector/blob/master/checklist.md)

dellachaise (approved):
Well done 👍

PR #900, opened 2022-09-12
Solution implemented

OlehMy:
[DEMO LINK](https://OlehMy.github.io/react_goods-selector/)


oksana-logos-frontend (approved):
Good solution))Your test  failed because you don't use RemoveButton.

src/App.tsx
```diff
@@ -15,71 +16,83 @@ export const goods = [
...
+                  ) : (
+                    <button
+                      data-cy="AddButton"
+                      type="button"
+                      onClick={() => addGood(good)}
+                      className="button"
+                    >
+                      +
+                    </button>
+                  )}
```
dellachaise, line 85: try to rewrite it using one button and change `- to +`, and need to add changes to click handler function.

dellachaise (changes requested):
Good job 👍 

PR #908, opened 2022-09-12
solution 1

kostya-kosyuk:
[DEMO LINK](https://kostya-kosyuk.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,89 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string | null;
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    selectedGood: 'Jam',
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  render() {
+    const { selectedGood } = this.state;
 
-    <table className="table">
-      <tbody>
-        <tr data-cy="Good">
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +15,89 @@ export const goods = [
...
+                <td>
+                  {
+                    good === selectedGood
+                      ? (
+                        <button
+                          data-cy="RemoveButton"
+                          type="button"
+                          className="button is-info"
+                          onClick={() => {
+                            this.setState({ selectedGood: null });
```
max-kravchenko, line 72: Checklist #2 
(Create separate methods instead of creating inline event handlers)

src/App.tsx
```diff
@@ -15,71 +15,89 @@ export const goods = [
...
+                          -
+                        </button>
+                      )
+                      : (
+                        <button
+                          data-cy="AddButton"
+                          type="button"
+                          className="button"
+                          onClick={() => {
+                            this.setState({ selectedGood: good });
```
max-kravchenko, line 84: Checklist #2 
(Create separate methods instead of creating inline event handlers)

max-kravchenko (approved):
Good job!
p.s. According to the checklist, you should create separate methods for event handling)

src/App.tsx
dellachaise: Agreed.

src/App.tsx
dellachaise: Agreed.

src/App.tsx
```diff
@@ -15,71 +15,89 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string | null;
```
dellachaise, line 19: Do you really need `null` here? Better set it in `State` to empty string `''`.

dellachaise (changes requested):
Great start 👍 

Just add changes according to comments and pay attention to [checklist point 2](https://github.com/mate-academy/react_goods-selector/blob/master/checklist.md).

andrii-dmytruk (approved):
Good job!

AndriiRut (approved):
Good job!

PR #912, opened 2022-09-13
Create React Goods Selector

max-kravchenko:
[DEMO LINK](https://max-kravchenko.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,92 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +16,92 @@ export const goods = [
...
+                        type="button"
+                        className="button"
+                        onClick={() => {
+                          this.selectGood(good);
+                        }}
+                      >
+                        +
+                      </button>
+
+                    )}
```
dellachaise, line 94: Try to create one button instead of 2.

dellachaise (approved):
Great job 👍

src/App.tsx
max-kravchenko: Thanks) Will try!

PR #917, opened 2022-09-13
Goods selector solution

ibondrk:
[DEMO LINK](https://ibondrk.github.io/react_goods-selector/)


Alzay007 (approved):
keep it up 👍

src/App.tsx
```diff
@@ -15,71 +16,92 @@ export const goods = [
...
-            Jam
-          </td>
-        </tr>
+        <table className="table">
+          <tbody>
+            {goods.map(good => {
+              const isSelected = selectedGood === good;
+              const goodItemBackground = classNames(
+                { 'has-background-success-light': isSelected },
+              );
```
denys-danyliuk, line 62: to me it looks like creating a variable here is unnecessary and waste of memory, but if some coach/mentor in your group told you to do that than it's fine, but I would like to hear explanation, and I see this stuff not only in your work, but in other students too so I'm just curious at this point

denys-danyliuk (approved):

src/App.tsx
ibondrk: I just used a variable here to improve the readability of the code. I agree with you, it's unnecessary at all, especially in this task, but I thought it was better to use a variable once instead of use condition every time in code.

PR #922, opened 2022-09-13
List added

Mejeezz:
[[DEMO LINK]](https://Mejeezz.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,9 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
-export const goods = [
+export const goods: string[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +16,94 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string | null;
```
ibondrk, line 20: I think you can't have `null` here. You had `''` or  some `good`

src/App.tsx
```diff
@@ -15,71 +16,94 @@ export const goods = [
...
-          </td>
+    const clearTitle = () => {
+      this.setState({ selectedGood: '' });
+    };
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+    const removeGood = (good: string) => (
```
ibondrk, line 35: Probably you should change name of this function becouse here you switch `selectGood` not only remove it

src/App.tsx
```diff
@@ -15,71 +16,94 @@ export const goods = [
...
-          </td>
+        <table className="table">
+          {goods.map(good => {
+            return (
+              <tbody>
+                <tr
+                  key={good}
+                  data-cy="Good"
+                  className={classNames(
+                    '', {
```
ibondrk, line 69: i think will be better if you move `{` to new line

src/App.tsx
```diff
@@ -15,71 +16,94 @@ export const goods = [
...
+                      'has-background-success-light': selectedGood === good,
+                    },
+                  )}
+                >
+                  <td>
+                    <button
+                      data-cy={
+                        selectedGood === good
```
ibondrk, line 78: if you want, you can create separate variable for this condition, as you using it's few times in your code

ibondrk (changes requested):
Good job! Change your function name `removeGood`. I think name don't describe what you realy do inside, and a few little changes of code readability

ibondrk (approved):
Well done🔥

src/App.tsx
```diff
@@ -15,71 +16,101 @@ export const goods = [
...
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+    const clearTitle = () => {
+      this.setState({ selectedGood: '' });
+    };
```
denys-danyliuk, line 33: move it outside the render method, otherwise it will be creating every time  when your component will render

src/App.tsx
```diff
@@ -15,71 +16,101 @@ export const goods = [
...
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+    const removeOrSelectGood = (good: string) => (
+      selectedGood === good
+        ? this.setState({ selectedGood: '' })
+        : this.setState({ selectedGood: good })
+    );
```
denys-danyliuk, line 39: same here

denys-danyliuk (changes requested):
Good job💪
Pay attention to the comments

denys-danyliuk (approved):
GJ!

PR #926, opened 2022-09-13
Solution of react_goods-selector

OleksiiPustovit:
[DEMO LINK](https://oleksiipustovit.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +17,98 @@ export const goods = [
...
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  isSelected = (good: string) => {
```
Desertiks, line 29: Better rename these methods like `handleIsSelected` or smth like that

src/App.tsx
```diff
@@ -15,71 +17,98 @@ export const goods = [
...
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string;
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
```
Desertiks, line 24:
```suggestion
type Props = {};
export class App extends React.Component<Props, State> {
```
It's better to write type `Props` even when it's empty

Desertiks (approved):
Nice work 🔥

denys-danyliuk (approved):
GJ!

PR #930, opened 2022-09-14
Solution

Black3942:


src/App.tsx
```diff
@@ -15,71 +16,83 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string | null;
```
alexandra-protyanova, line 20: It probably better to use empty string instead of null

alexandra-protyanova (approved):

PR #943, opened 2022-09-15
react goods selector

erikkaraman:


Anatolii-Bukoros (approved):
good job

dellachaise (approved):
Good job 👍 

But [checklist point 1](https://github.com/mate-academy/react_goods-selector/blob/master/checklist.md) still actual.
Method's name should start with a **verb** (so you could easily tell what your method actually do)
`handlerUnselectGood` and `handlerSelect` would be better.

PR #949, opened 2022-09-17
solution

Denisdkoval:
https://denisdkoval.github.io/react_goods-selector/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import { Component } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,92 @@ export const goods = [
...
+                      <button
+                        data-cy="AddButton"
+                        type="button"
+                        className="button"
+                        onClick={() => {
+                          this.toggleSelection(good);
+                        }}
+                      >
+                        +
+                      </button>
```
SanyaBratashchuk, line 94: You can write ternary inside a button to avoid duplication

SanyaBratashchuk (changes requested):

src/App.tsx
Denisdkoval: I don't think this is the right decision. This will spoil the readability of the code, these buttons have only 'type' in common

SanyaBratashchuk: onClick too

SanyaBratashchuk: also have a common classname

SanyaBratashchuk (changes requested):
⬆️ conversation

src/App.tsx
Denisdkoval: You're joking? do they have different classes, or should an additional class also be added through a ternary operatop?

Denisdkoval: ok can you show whot do you mind?

PR #956, opened 2022-09-25
solution

HusakOleh:
[goods selector](https://HusakOleh.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


Reptiloid044 (approved):
Nice solution!

src/App.tsx
```diff
@@ -15,71 +15,91 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+export class App extends Component<{}, { selectedGood: string }> {
```
dellachaise, line 18: Create separate `type State` and use it here.

src/App.tsx
```diff
@@ -15,71 +15,91 @@ export const goods = [
...
-          </td>
-        </tr>
+    return (
+      <main className="section container">
+        {selectedGood
+          ? (
+            <h1 className="title is-flex is-align-items-center">
+              {selectedGood}
+              {' '}
+              {' is selected'}
```
dellachaise, line 41: Use [interpolation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) here.

src/App.tsx
```diff
@@ -15,71 +15,91 @@ export const goods = [
...
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+              <button
+                data-cy="ClearButton"
+                type="button"
+                className="delete ml-3"
+                onClick={this.clearState}
+              />
+            </h1>
+          )
+          : <h1 className="title">No goods selected</h1>}
```
dellachaise, line 52: And no need to write 2 `h1` you could use ternary operator inside `h1`
`<h1>ternary operator </h1>`

dellachaise (changes requested):
Good start 👍 

Just add changes according to commens 👇

src/App.tsx
```diff
@@ -15,71 +15,96 @@ export const goods = [
...
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string,
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends Component<{}, State> {
+  state: { selectedGood: string } = {
```
dellachaise, line 23:
```suggestion
  state: Readonly<State> = {
```

dellachaise (approved):
Well done 👍

PR #959, opened 2022-09-28
Solution

ElinaBahirova:
https://elinabahirova.github.io/react_goods-selector/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,87 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string;
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state = {
+    selectedGood: 'Jam',
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +15,87 @@ export const goods = [
...
-            >
-              -
-            </button>
-          </td>
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
+                data-cy="Good"
+                key={goods.indexOf(good)}
```
SanyaBratashchuk, line 56:
```suggestion
                key={good}
```
Index it is default value for key

src/App.tsx
```diff
@@ -15,71 +15,87 @@ export const goods = [
...
+                className={
+                  selectedGood === good
+                    ? 'has-background-success-light'
+                    : ''
```
SanyaBratashchuk, line 60: use classnames library

SanyaBratashchuk (changes requested):

src/App.tsx
```diff
@@ -15,71 +16,85 @@ export const goods = [
...
+              >
+                <td>
+                  {selectedGood !== good
+                    ? (
+                      <button
+                        data-cy="AddButton"
+                        type="button"
+                        className="button"
+                        onClick={() => {
+                          this.setState({ selectedGood: good });
```
liliia-livandovskaya, line 70: it would be better to add handlers on each button click

liliia-livandovskaya (approved):
GJ 🔥

PR #960, opened 2022-09-29
Solution

PaulMaistrenko:
[DEMO LINK](https://PaulMaistrenko.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,7 @@
-import React from 'react';
+/* eslint-disable jsx-a11y/control-has-associated-label */
+/* eslint-disable max-len */
+/* eslint-disable react/prefer-stateless-function */
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +18,84 @@ export const goods = [
...
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  handlerAddGood = (good: string) => this.setState({ selectedGood: good });
```
SanyaBratashchuk, line 30:
```suggestion
  handleAddGood = (good: string) => this.setState({ selectedGood: good });
```

SanyaBratashchuk: But why cant you do just `this.setState({selectedGood:good})` what is the purpose of this function

src/App.tsx
```diff
@@ -15,71 +18,84 @@ export const goods = [
...
+                key={good}
+                className={(selectedGood === good) ? 'has-background-success-light' : ''}
+              >
+                {selectedGood === good
+                  ? (
+                    <td>
+                      <button
+                        data-cy="RemoveButton"
+                        type="button"
+                        className={(selectedGood === good) ? 'button is-info' : 'button'}
```
SanyaBratashchuk, line 69: use classnames library

SanyaBratashchuk (changes requested):

src/App.tsx
PaulMaistrenko: it was corrected

src/App.tsx
TarasHoliuk: :arrow_up: it is already installed by `npm`, you need just import it

src/App.tsx
```diff
@@ -1,4 +1,7 @@
-import React from 'react';
+/* eslint-disable jsx-a11y/control-has-associated-label */
+/* eslint-disable max-len */
+/* eslint-disable react/prefer-stateless-function */
```
TarasHoliuk, line 3: should work without these rules
feel free to ask in fe_chat if you need help with it

TarasHoliuk (changes requested):

src/App.tsx
```diff
@@ -15,71 +17,90 @@ export const goods = [
...
-  </main>
-);
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
+                data-cy="Good"
+                key={good}
+                className={
+                  classNames('', {
```
TarasHoliuk, line 61:
```suggestion
                  classNames({
```

src/App.tsx
```diff
@@ -15,71 +17,90 @@ export const goods = [
...
+    } = this.state;
 
-        <tr data-cy="Good">
-          <td>
+    return (
+      <main className="section container">
+        <h1
+          className={selectedGood
+            ? 'title is-flex is-align-items-center'
+            : 'title'}
```
TarasHoliuk, line 41: use classNames here as well

TarasHoliuk (changes requested):

src/App.tsx
PaulMaistrenko: fixed

marina-tilniak (approved):

PR #970, opened 2022-10-06
solution 

fatihsamur:
[DEMO LINK](https://fatihsamur.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+/* eslint-disable */
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,4 +1,5 @@
-import React from 'react';
+/* eslint-disable */
```
kamal-tufekcic, line 1: This is a really bad idea

kamal-tufekcic (approved):
Don't think they'll be too happy about using `e` as argument name

src/App.tsx
fatihsamur: omg I should have deleted this before pushing it :D I disabled it for not to be annoyed while dealing with other stuff but it looks it is not a good idea :D

src/App.tsx
```diff
@@ -15,71 +15,90 @@ export const goods = [
...
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+export const App: React.FC = () => {
+  const defaultGood = 'Jam';
+  const defaultSelection = goods.indexOf(defaultGood);
+  const [selectedGood, setSelectedGood] = useState(defaultGood);
+  const [selectedTr, setSelectedTr] = useState<string | null>(defaultSelection
```
ihor-jpeg, line 22: not quite clear what selectedTr means
use more meaningful names

src/App.tsx
```diff
@@ -15,71 +15,90 @@ export const goods = [
...
+    .toString());
+  const AddSelectHandler = (e: React.MouseEvent, good: string) => {
```
ihor-jpeg, line 24:
```suggestion
  const AddSelectHandler = (
    event: React.MouseEvent,
    good: string,
  ) => {
```

src/App.tsx
```diff
@@ -15,71 +15,90 @@ export const goods = [
...
+                  <button
+                    data-cy="AddButton"
+                    type="button"
+                    className="button"
+                    onClick={(e) => AddSelectHandler(e, good)}
+                  >
+                    +
+                  </button>
+                </td>
+              )}
```
ihor-jpeg, line 93: using logical && would make this a lot more readable

ihor-jpeg (changes requested):

src/App.tsx
```diff
@@ -15,71 +15,100 @@ export const goods = [
...
-          </td>
-        </tr>
+  return (
+    <main className="section container">
+      <h1 className="title is-flex is-align-items-center">
+        {selectedGood ? (
+          `${selectedGood} is selected`
+        ) : (
+          'No goods selected'
+        )}
```
ihor-jpeg, line 54: make sure that you don't have complex structures in the return statement
you can create a variable before the return and simply use it

ihor-jpeg (commented):
also, why do you convert indexes to strings?
i think you can do it without .toString method

ihor-jpeg (changes requested):
left a few more comments

ihor-jpeg (approved):

PR #975, opened 2022-10-13
goods selector created

illia-sosnytskyi:
[DEMO LINK](https://illia-sosnytskyi.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,103 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selected: boolean,
+  selectedGood: string,
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state = {
+    selected: true,
+    selectedGood: 'Jam',
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  pickGood = (good: string) => {
+    this.setState({
+      selected: true,
+      selectedGood: good,
+    });
+  };
 
-    <table className="table">
-      <tbody>
-        <tr data-cy="Good">
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +15,103 @@ export const goods = [
...
-  </main>
-);
+        <table className="table">
+          <tbody>
+            {
+              goods.map(good => (
+                <tr
+                  data-cy="Good"
+                  key={good}
+                  className={this.classForGood(selectedGood, good)}
```
myroslav-diiak, line 79: i think classNames will be better here

myroslav-diiak (approved):

src/App.tsx
OleksiiBlezniuk: use classNames 

src/App.tsx
```diff
@@ -15,71 +15,103 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selected: boolean,
```
OleksiiBlezniuk, line 19: why do you need this state?

src/App.tsx
```diff
@@ -15,71 +15,103 @@ export const goods = [
...
+                {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+                <button
+                  data-cy="ClearButton"
+                  type="button"
+                  className="delete ml-3"
+                  onClick={this.removeGood}
+                />
+              </h1>
+            )
+            : <h1 className="title">No goods selected</h1>
```
OleksiiBlezniuk, line 69: In both cases it will be `h1`, no need to duplicate it in ternary operator, just select needed classes with `classnames`

OleksiiBlezniuk (changes requested):

src/App.tsx
```diff
@@ -15,71 +16,99 @@ export const goods = [
...
+            {selectedGood ? `${selectedGood} is selected` : 'No goods selected'}
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Jam
-          </td>
-        </tr>
+            {
+              selectedGood
+              && (
+                /* eslint-disable-next-line jsx-a11y/control-has-associated-label */
```
OleksiiBlezniuk, line 55: better try to fix the issue instead of disable it via comments

OleksiiBlezniuk (approved):

PR #978, opened 2022-10-13
solution

vmamchur:
[DEMO LINK](https://vmamchur.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,95 @@ export const goods = [
...
+                data-cy="ClearButton"
+                type="button"
+                className="delete ml-3"
+                onClick={() => {
+                  this.setState(removeGood);
+                }}
+              />
+            </h1>
+          )
+          : (<h1 className="title">No goods selected</h1>)}
```
imaksweb, line 66: It seems like repeating code here, in both cases you have `h1`, differs only classes and content

src/App.tsx
```diff
@@ -15,71 +16,95 @@ export const goods = [
...
+                      <button
+                        data-cy="AddButton"
+                        type="button"
+                        className="button"
+                        onClick={() => addGood(good)}
+                      >
+                        +
+                      </button>
+                    )}
+                </td>
```
imaksweb, line 98: Also repeating code

imaksweb (approved):
Nice work 👍 
Just a few suggestions but I'm not sure you should follow them 🙂

src/App.tsx
vmamchur: good point. i'll fix it later
thx!

src/App.tsx
```diff
@@ -15,71 +16,96 @@ export const goods = [
...
-            >
-              +
-            </button>
-          </td>
+              className="delete ml-3"
+              onClick={() => {
+                this.setState(removeGood);
+              }}
+            />
+          )}
```
silchyn, line 70: You already have checked for 'selectedGood' in the ternary, so there is no reason to repeat that

src/App.tsx
```diff
@@ -15,71 +16,96 @@ export const goods = [
...
+        </h1>
+
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
+                key={good}
```
silchyn, line 77: Maybe we should use another identifier. The name of a good is not guaranteed to be unique

silchyn (changes requested):

silchyn (approved):
Cool!

PR #979, opened 2022-10-13
goods selector

Shepeha-Andriy:
https://Shepeha-Andriy.github.io/react_goods-selector/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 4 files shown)


Veremchukk (commented):
Well done

src/App.tsx
```diff
@@ -15,71 +16,87 @@ export const goods = [
...
+                    {
+                      'has-background-success-light':
+                      this.state.selectedWord === good,
+                    },
+                  )}
+                >
+                  <td>
+                    <button
+                      data-cy={
+                        this.state.selectedWord === good
```
OleksiiBlezniuk, line 66: you've used this check 5 times, let's move it to constant
also would be nice to create constants for buttons text, callback, data-cy instead of using ternary operators everywhere, but this point is less critical

OleksiiBlezniuk (changes requested):

src/App.tsx
```diff
@@ -15,71 +16,95 @@ export const goods = [
...
+    return (
+      <main className="section container">
+        {
+          !selectedGood
+            ? (
+              <h1 className="title">No goods selected</h1>)
+            : (
+              <h1 className="title is-flex is-align-items-center">
+                {`${selectedGood} `}
+                is selected
```
OleksiiBlezniuk, line 44:
```suggestion
                {`${selectedGood} is selected`}
```

OleksiiBlezniuk: no need to start it at new line

OleksiiBlezniuk (approved):

PR #981, opened 2022-10-14
feat: goods selector component was implemented

grafit-off:
[DEMO LINK](https://grafit-off.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import { Component } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,90 @@ export const goods = [
...
+                  this.setState({ selectedGood: '' });
+                }}
+              />
+            </h1>
+          )
+          : (
+            <h1 className="title">
+              No goods selected
+            </h1>
+          )}
```
imaksweb, line 53: Looks like repeating code here, in both cases you have `h1` just with different classes and content

src/App.tsx
```diff
@@ -15,71 +16,90 @@ export const goods = [
...
+                        data-cy="AddButton"
+                        type="button"
+                        className="button"
+                        onClick={() => {
+                          this.setState({ selectedGood: item });
+                        }}
+                      >
+                        +
+                      </button>
+                    )}
```
imaksweb, line 92: The same thing here

imaksweb (approved):
GJ 👍 
Just a few suggestions ⬇️

Anna-Tkachenko (approved):
Good job!

PR #982, opened 2022-10-14
Implement goods selection

imaksweb:
- [DEMO LINK](https://imaksweb.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import { Component } from 'react';
+import cn from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


vmamchur (approved):
Cool 🔥

src/App.tsx
```diff
@@ -15,71 +16,106 @@ export const goods = [
...
-          </td>
-        </tr>
-      </tbody>
-    </table>
-  </main>
-);
+              return (
+                <tr
+                  data-cy="Good"
+                  key={good}
```
silchyn, line 83: Maybe we should use another identifier or create one? The name of a good is not guaranteed to be unique

silchyn (changes requested):

silchyn (changes requested):
Do not use index as key (checklist), try using the 'uuid' package

imaksweb:
> Do not use index as key (checklist), try using the 'uuid' package

Indeed, completely forgot about it

silchyn (approved):
Cool!

PR #984, opened 2022-10-14
Implemented goods selector

Andrew-Skoryk:
[DEMO LINK](https://Andrew-Skoryk.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,85 +1,109 @@
-import React from 'react';
+import { Component, MouseEvent } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
-  'Dumplings',
-  'Carrot',
-  'Eggs',
-  'Ice cream',
-  'Apple',
-  'Bread',
-  'Fish',
-  'Honey',
-  'Jam',
-  'Garlic',
+  { name: 'Dumplings', id: 1 },
+  { name: 'Carrot', id: 2 },
+  { name: 'Eggs', id: 3 },
+  { name: 'Ice cream', id: 4 },
+  { name: 'Apple', id: 5 },
+  { name: 'Bread', id: 6 },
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,85 +1,109 @@
...
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Jam
-          </td>
-        </tr>
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+              <button
+                data-cy="ClearButton"
+                type="button"
+                className="delete ml-3"
+                onClick={this.removeOnClick}
```
xiimiik, line 51: i think we can change state without creating another func, same below

src/App.tsx
```diff
@@ -1,85 +1,109 @@
...
+                    {isSelected
+                      ? (
+                        <button
+                          data-cy="RemoveButton"
+                          type="button"
+                          className="button is-info"
+                          onClick={this.removeOnClick}
+                        >
+                          -
+                        </button>
```
xiimiik, line 82: Try to change `<button>` tag, not create another one

xiimiik (changes requested):
GJ, but need some refactor

src/App.tsx
Andrew-Skoryk: Isn't this a good practice to separate code?

src/App.tsx
```diff
@@ -1,85 +1,105 @@
...
+                  data-cy="Good"
+                  className={classNames({
+                    'has-background-success-light': isSelected,
+                  })}
+                >
+                  <td>
+                    <button
+                      data-cy={isSelected
+                        ? 'RemoveButton'
+                        : 'AddButton'}
```
OleksiiBlezniuk, line 74: Or you could save this value in variable and pass it here

OleksiiBlezniuk (approved):

PR #989, opened 2022-10-16
Develop

vladimir-dublya:
[DEMO LINK](https://vladimir-dublya.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,85 +1,89 @@
-import React from 'react';
-import 'bulma/css/bulma.css';
-import './App.scss';
+import React from "react";
+import "bulma/css/bulma.css";
+import "./App.scss";
 
 export const goods = [
-  'Dumplings',
-  'Carrot',
-  'Eggs',
-  'Ice cream',
-  'Apple',
-  'Bread',
-  'Fish',
-  'Honey',
-  'Jam',
-  'Garlic',
+  { id: 1, name: "Dumplings" },
+  { id: 2, name: "Carrot" },
+  { id: 3, name: "Eggs" },
+  { id: 4, name: "Ice cream" },
+  { id: 5, name: "Apple" },
+  { id: 6, name: "Bread" },
+  { id: 7, name: "Fish" },
+  { id: 8, name: "Honey" },
+  { id: 9, name: "Jam" },
+  { id: 10, name: "Garlic" },
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string;
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+type Props = {};
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+export class App extends React.Component<Props, State> {
+  state: Readonly<State> = {
+    // eslint-disable-next-line react/no-unused-state
+    selectedGood: "",
+  };
 
-    <table className="table">
-      <tbody>
-        <tr data-cy="Good">
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+  render(): React.ReactNode {
+    return (
+      <main className="section container">
+        <h1 className="title is-flex is-align-items-center">
+          {this.state.selectedGood
+            ? `${this.state.selectedGood} is selected`
+            : "No goods selected"}
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+          <button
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -3,83 +3,87 @@ import 'bulma/css/bulma.css';
...
+  { id: 2, name: 'Carrot' },
+  { id: 3, name: 'Eggs' },
+  { id: 4, name: 'Ice cream' },
+  { id: 5, name: 'Apple' },
+  { id: 6, name: 'Bread' },
+  { id: 7, name: 'Fish' },
+  { id: 8, name: 'Honey' },
+  { id: 9, name: 'Jam' },
+  { id: 10, name: 'Garlic' },
 ];
```
nikitaovch99, line 16: key needs a unique identificator(but not the index) but you don't have to create id for it specifically because you already have a unique identificator in this task = name it's unique and you can use it as a key.

src/App.tsx
```diff
@@ -3,83 +3,87 @@ import 'bulma/css/bulma.css';
...
+                  <button
+                    data-cy="AddButton"
+                    type="button"
+                    className={`button ${
+                      item.name === this.state.selectedGood ? 'is-info' : ''
+                    }`}
+                    onClick={() => {
+                      // eslint-disable-next-line react/no-unused-state
+                      this.setState(
+                        // eslint-disable-next-line react/no-access-state-in-setstate
```
nikitaovch99, line 68: You could just write setState in IF if your selectedgood === item.name and eslint problems would disappear

nikitaovch99 (changes requested):

src/App.tsx
nikitaovch99: Good job for removing id, but I also meant you don't even need to create an array of objects just use already given array. But whatever floats your boat I guess.

nikitaovch99 (approved):

src/App.tsx
Anna-Tkachenko: check @nikitaovch99 comment
you don't need create an array of objects here, it's redundant

src/App.tsx
```diff
@@ -3,83 +3,85 @@ import 'bulma/css/bulma.css';
...
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+export class App extends React.Component<Props, State> {
+  state: Readonly<State> = {
+    // eslint-disable-next-line react/no-unused-state
```
Anna-Tkachenko, line 26: delete this comment

src/App.tsx
```diff
@@ -3,83 +3,85 @@ import 'bulma/css/bulma.css';
...
-              className="button"
-            >
-              +
-            </button>
-          </td>
+  render(): React.ReactNode {
+    return (
+      <main className="section container">
+        <h1 className="title is-flex is-align-items-center">
+          {this.state.selectedGood
```
Anna-Tkachenko, line 34: use destructuring for getting needed props from the state object

src/App.tsx
```diff
@@ -3,83 +3,85 @@ import 'bulma/css/bulma.css';
...
+          <tbody>
+            {goods.map((item) => (
+              <tr
+                data-cy="Good"
+                key={item.name}
+                className={`${
+                  item.name === this.state.selectedGood
+                    ? 'has-background-success-light'
+                    : ''
+                }`}
```
Anna-Tkachenko, line 56: use classNames library

src/App.tsx
```diff
@@ -3,83 +3,85 @@ import 'bulma/css/bulma.css';
...
+              >
+                <td>
+                  <button
+                    data-cy="AddButton"
+                    type="button"
+                    className={`button ${
+                      item.name === this.state.selectedGood ? 'is-info' : ''
```
Anna-Tkachenko, line 63: same here

src/App.tsx
```diff
@@ -3,83 +3,85 @@ import 'bulma/css/bulma.css';
...
+                    }`}
+                    onClick={() => {
+                      if (this.state.selectedGood === item.name) {
+                        this.setState({ selectedGood: '' });
+                      } else {
+                        this.setState({ selectedGood: item.name });
+                      }
+                    }}
+                  >
+                    {item.name === this.state.selectedGood ? '-' : '+'}
```
Anna-Tkachenko, line 73: you repeat this check many times, move it to separate function or create separate boolean variable for it

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -15,71 +16,88 @@ export const goods = [
...
-          <td>
-            <button
-              data-cy="RemoveButton"
-              type="button"
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+  render(): React.ReactNode {
```
Denchepornyuk, line 38:
```suggestion
  render() {
```

Denchepornyuk (approved):

PR #992, opened 2022-10-16
Implemented App component which render goods list with the ability to select one or clear selection

Dikey945:
[DEMO LINK](https://Dikey945.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,88 @@ export const goods = [
...
+                        <button
+                          data-cy="RemoveButton"
+                          type="button"
+                          className="button is-info"
+                          onClick={this.removeSelection}
+                        >
+                          -
+                        </button>
+                      </td>
+                    ) : (
```
Yevheniya-Deryka, line 78: probably, there must be a new line before :

src/App.tsx
```diff
@@ -15,71 +16,88 @@ export const goods = [
...
+                      <td>
+                        <button
+                          data-cy="AddButton"
+                          type="button"
+                          className="button"
+                          onClick={() => (
```
Yevheniya-Deryka, line 84: the check-list tells it's better to move such functions to the separate constant, as far as I've understood

Yevheniya-Deryka (approved):
GJ! 
Left you a couple of comments

src/App.tsx
```diff
@@ -15,71 +16,88 @@ export const goods = [
...
-              type="button"
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+        {!selectedGood
+          ? <h1 className="title">No goods selected</h1>
+          : (
+            <h1 className="title is-flex is-align-items-center">
```
silchyn, line 41: h1 repeats, only className differs

src/App.tsx
```diff
@@ -15,71 +16,88 @@ export const goods = [
...
-            Garlic
-          </td>
-        </tr>
-      </tbody>
-    </table>
-  </main>
-);
+              return (
+                <tr
+                  key={good}
```
silchyn, line 62: Maybe we should use another identifier? The name of a good is not guaranteed to be unique

silchyn (changes requested):
And points above

src/App.tsx
Dikey945: Which identifier should I take from the array if the array index prohibited to use by checklist?

silchyn (changes requested):
Re-read the comments

silchyn (approved):
My bad

Andr1yK (changes requested):
test

Andr1yK (approved):

Andr1yK (changes requested):
test

Andr1yK (approved):

PR #993, opened 2022-10-17
Solution

Proph7000:
[DEMO LINK](https://proph7000.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,131 @@ export const goods = [
...
+}
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    selectedGood: 'Jam',
+    isSelected: true,
```
tetiana-ivanchenko, line 43: why do you need this "isSelected" property? you can just use the selectedGood.
if this is not an empty string, thats mean that some Good is selected

tetiana-ivanchenko: this will simplify your code and the methods you created

tetiana-ivanchenko (approved):

src/App.tsx
tetiana-ivanchenko, line 43: why do you need this "isSelected" property? you can just use the selectedGood.
if this is not an empty string, thats mean that some Good is selected


this will simplify your code and the methods you created

tetiana-ivanchenko (changes requested):

nazarmatsevych (approved):

PR #997, opened 2022-10-17
The goods selector was implemented

anastasiia-nahaiska:
[DEMO LINK](https://anastasiia-nahaiska.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


ArturNagaycev (approved):
👍

src/App.tsx
```diff
@@ -15,71 +16,85 @@ export const goods = [
...
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+    return (
+      <main className="section container">
+        {selectedGood
+          ? (
+            <h1 className="title is-flex is-align-items-center">
```
silchyn, line 43: h1 repeats, only className differs

src/App.tsx
```diff
@@ -15,71 +16,85 @@ export const goods = [
...
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Jam
-          </td>
-        </tr>
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
+                data-cy="Good"
+                key={good}
```
silchyn, line 60: Maybe we should use another identifier? The name of a good is not guaranteed to be unique

src/App.tsx
```diff
@@ -15,71 +16,85 @@ export const goods = [
...
+                      </button>
+                    ) : (
+                      <button
+                        data-cy="AddButton"
+                        type="button"
+                        className="button"
+                        onClick={() => {
+                          this.saveGood(good);
+                        }}
+                      >
```
silchyn, line 84: Button itself repeats, only some properties differ

silchyn (changes requested):

silchyn (changes requested):
Good job, almost done

PR #998, opened 2022-10-17
React component with goods-selector was created

Yevheniya-Deryka:
[DEMO LINK](https://Yevheniya-Deryka.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,14 @@
-import React from 'react';
+import { Component } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
-export const goods = [
+type Good = string;
+type State = {
+  selectedGood: Good | null;
+};
+
+export const goods: Good[] = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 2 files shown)


Dikey945:
Good Job 🔥

src/App.tsx
```diff
@@ -1,8 +1,14 @@
-import React from 'react';
+import { Component } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
-export const goods = [
+type Good = string;
+type State = {
+  selectedGood: Good | null;
```
Anna-Tkachenko, line 8: will be better to store empty string here instead of null

src/App.tsx
```diff
@@ -15,71 +21,82 @@ export const goods = [
...
+              const isSelected = selectedGood === good;
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Jam
-          </td>
-        </tr>
+              const handleSelection = () => (
+                isSelected
+                  ? this.setState({ selectedGood: '' })
+                  : this.setState({ selectedGood: good })
```
Anna-Tkachenko, line 62: will be better specify needed argument to the selectedGood value according to `isSelected`

Anna-Tkachenko (approved):

PR #999, opened 2022-10-17
goods selector implemented

artem-musii:
[DEMO LINK](https://artem-musii.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +16,92 @@ export const goods = [
...
+  };
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+  render() {
+    return (
+      <main className="section container">
+        {this.state.selectedGood
```
silchyn, line 39: Destructure the state

src/App.tsx
```diff
@@ -15,71 +16,92 @@ export const goods = [
...
+          ? (
+            <h1 className="title is-flex is-align-items-center">
```
silchyn, line 41: h1 repeats, className differs

src/App.tsx
```diff
@@ -15,71 +16,92 @@ export const goods = [
...
-            >
-              -
-            </button>
-          </td>
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
+                data-cy="Good"
+                key={good}
```
silchyn, line 63: Maybe we should use another identifier? The name of a good is not guaranteed to be unique

silchyn (changes requested):

src/App.tsx
```diff
@@ -15,71 +17,85 @@ export const goods = [
...
+          {selectedGood
+            ? (
+              <>
+                {`${selectedGood} is selected`}
+                {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+                <button
+                  data-cy="ClearButton"
+                  type="button"
+                  className="delete ml-3"
+                  onClick={() => this.toggleSelected(undefined)}
```
silchyn, line 52:
```suggestion
                  onClick={() => this.toggleSelected()}
```

src/App.tsx
```diff
@@ -15,71 +17,85 @@ export const goods = [
...
+                className={classNames(
+                  {
+                    'has-background-success-light':
+                    good === selectedGood,
+                  },
+                )}
+              >
+                <td>
+                  <button
+                    data-cy={selectedGood === good
```
silchyn, line 74: `selectedGood === good` repeats several times, so we should create variable

silchyn (changes requested):

PR #1000, opened 2022-10-17
added goods table,interactive clicks and logic

SerhiiRomaniukWD:
[DEMO LINK](https://serhiiromaniukwd.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -59,12 +59,12 @@ export class App extends Component<{}, State> {
 
         <table className="table">
           <tbody>
-            {goods.map((good) => {
+            {goods.map((good, index) => {
               const goodSelected = selectedGood === good;
 
               return (
                 <tr
-                  key={good}
+                  key={good + index.toString()}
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,71 +16,95 @@ export const goods = [
...
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+  removeGood = () => (
```
vladyslav-harkusha, line 32: maybe you should rename methods to "handleRemoveGood" or "removeGoodHandler"

vladyslav-harkusha (approved):

src/App.tsx
SerhiiRomaniukWD: I think 'handler' word can't explain what this function do, but you are right I should rename this method

OleksiiBlezniuk (approved):

PR #1009, opened 2022-10-18
Rendered goods in a table with the ability to select one or clear sel…

OleksandrGlulhovskiy:
[DEMO LINK](https://OleksandrGlulhovskiy.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,96 @@ export const goods = [
...
-          </td>
-        </tr>
-      </tbody>
-    </table>
-  </main>
-);
+              return (
+                <tr
+                  data-cy="Good"
+                  key={good}
```
SerhiiRomaniukWD, line 71: array can contain the same elements
it will be better combine good & index
key={good + index.toString()}

src/App.tsx
```diff
@@ -15,71 +16,96 @@ export const goods = [
...
+                    : ''}
+                >
+                  <td>
+                    {isSelected
+                      ? (
+                        <button
+                          data-cy="RemoveButton"
+                          type="button"
+                          className="button is-info"
+                          onClick={this.removeGood}
```
SerhiiRomaniukWD, line 83: {this.removeGood} you use it more than one time
destructuring is your friend

SerhiiRomaniukWD (approved):
good job, but you must fix minor mistakes

src/App.tsx
Anna-Tkachenko: use classNames library for setting class according to some check

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #1012, opened 2022-10-18
React goods selector

glu3s:
[DEMO LINK](https://glugovsky.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -52,45 +52,49 @@ export class App extends React.Component<{}, State> {
 
         <table className="table">
           <tbody>
-            {goods.map(good => (
-              <tr
-                data-cy="Good"
-                key={good + goods.indexOf(good)}
-                className={
-                  classNames({
-                    'has-background-success-light': good === selectedGood,
-                  })
-                }
-              >
-                <td>
-                  {good === selectedGood
-                    ? (
-                      <button
-                        data-cy="RemoveButton"
-                        type="button"
-                        onClick={this.clearGood}
-                        className="button is-info"
-                      >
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +16,82 @@ export const goods = [
...
+                      classNames({
+                        button: true,
+                        'is-info': good === selectedGood,
+                      })
+                    }
+                  >
+                    {good === selectedGood
+                      ? '-'
+                      : '+'}
+                  </button>
```
jojnajbude, line 84: to avoid so many ternary operators, u can create separate button for each cases.

jojnajbude (approved):

src/App.tsx
```diff
@@ -15,71 +16,82 @@ export const goods = [
...
-        </tr>
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
+                data-cy="Good"
+                key={good + goods.indexOf(good)}
+                className={
+                  classNames({
+                    'has-background-success-light': good === selectedGood,
```
Anna-Tkachenko, line 61: you use the check if current good is selected several times
create boolean variable for those value or separate function with those logic

Anna-Tkachenko (changes requested):

src/App.tsx
Anna-Tkachenko: you didn't fix the comment

Anna-Tkachenko (changes requested):

Denchepornyuk (approved):

PR #1015, opened 2022-10-18
feat: add app component

iryna-borniak:
[DEMO LINK](https://iryna-borniak.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,11 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+
+type State = {
+  selectedGood: string;
+  title: string;
+};
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,7 +1,13 @@
-import React from 'react';
+import { Component } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
+type State = {
+  selectedGood: string;
+  title: string;
```
Anna-Tkachenko, line 8: title variable in state is redundant, delete it

src/App.tsx
```diff
@@ -15,71 +21,100 @@ export const goods = [
...
+      return selectedGood === item;
+    }
 
-        <tr data-cy="Good">
-          <td>
+    return (
+      <main className="section container">
+        {title ? (
+          <h1 className="title">{title}</h1>
+        ) : (
```
Anna-Tkachenko, line 65: you can check here selectedGood variable, and print needed text if there is no selected good
title variable is redundant

src/App.tsx
```diff
@@ -15,71 +21,100 @@ export const goods = [
...
+            />
+          </h1>
+        )}
+
+        <table className="table">
+          <tbody>
+            {goods.map((item: string, index: number) => (
+              <tr
+                data-cy="Good"
+                key={item + String(index)}
```
Anna-Tkachenko, line 83: string(index) is also redundant, use just simple item name here

Anna-Tkachenko (changes requested):

src/App.tsx
iryna-borniak: done

src/App.tsx
iryna-borniak: done

src/App.tsx
iryna-borniak: done

Anna-Tkachenko (approved):
⭐️

PR #1016, opened 2022-10-19
goods selector was implemented

dsydorenko534:
[DEMO LINK](https://dsydorenko534.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,4 @@
-import React from 'react';
-import classNames from 'classnames';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


iPanda177 (approved):
gj

src/App.tsx
```diff
@@ -15,71 +16,108 @@ export const goods = [
...
-            </button>
-          </td>
+  selectGood(good: string) {
+    const { selectedGood } = this.state;
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+    if (!selectedGood.length) {
```
OleksiiBlezniuk, line 36:
```suggestion
    if (!selectedGood) {
```

src/App.tsx
```diff
@@ -15,71 +16,108 @@ export const goods = [
...
+      this.setState({ selectedGood: good });
+    } else if (selectedGood !== good) {
+      this.setState({ selectedGood: good });
+    } else {
+      this.setState({ selectedGood: '' });
```
OleksiiBlezniuk, line 41: How can function that named as selectGood can remove good?
Better let's select needed callback function via ternary operator, that way will be more understandable

src/App.tsx
```diff
@@ -15,71 +16,108 @@ export const goods = [
...
-            Jam
-          </td>
-        </tr>
+    return (
+      <main className="section container">
+        <h1
+          className={
+            classNames(
+              'title',
+              { 'is-flex is-align-items-center': selectedGood.length > 0 },
```
OleksiiBlezniuk, line 54: Why do you need to check empty strings length, if its convertion to boolean will return false? 🙂

OleksiiBlezniuk, line 54:
```suggestion
              { 'is-flex is-align-items-center': !selectedGood },
```

src/App.tsx
```diff
@@ -15,71 +16,108 @@ export const goods = [
...
+        <table className="table">
+          <tbody>
+            {goods.map((good) => (
+              <tr
+                data-cy="Good"
+                key={good}
+                className={classNames(
+                  '',
+                  {
+                    'has-background-success-light': selectedGood === good,
```
OleksiiBlezniuk, line 85: let's save result of this check to constant, you've used it 4 times, no need to repeat yourself

OleksiiBlezniuk (changes requested):

src/App.tsx
```diff
@@ -15,71 +15,78 @@ export const goods = [
...
+                    <button
+                      data-cy="RemoveButton"
+                      type="button"
+                      className="button is-info"
+                      onClick={resetState}
+                    >
+                      -
+                    </button>
+                  </td>
+                )}
```
OleksiiBlezniuk, line 81: Your previous solution was good, it's nice to reuse elements and conditionally select values for its attributes, why did you decide to remade it like this? You could just fix comments that I left

src/App.tsx
```diff
@@ -15,71 +15,78 @@ export const goods = [
...
+              className={good === selectedGood
+                ? 'has-background-success-light'
+                : ''}
             >
-              +
-            </button>
-          </td>
+              {selectedGood !== good
+                ? (
+                  <td>
```
OleksiiBlezniuk, line 58: This element is duplicated, move it outside ternary operator

src/App.tsx
```diff
@@ -15,71 +15,78 @@ export const goods = [
...
-          </td>
-        </tr>
+              className="delete ml-3"
+              onClick={resetState}
+            />
+          </h1>
+        ) : (
+          <h1 className="title">
+            No goods selected
+          </h1>
```
OleksiiBlezniuk, line 43: I just left comment about string length check, why did you decide to duplicate this heading instead of previous solution with reusing it? (which were better)

src/App.tsx
```diff
@@ -15,71 +15,78 @@ export const goods = [
...
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+export const App: React.FC = () => {
+  const [selectedGood, setSelectedGood] = useState<string | null>('Jam');
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+  const resetState = () => {
+    setSelectedGood(null);
+  };
```
OleksiiBlezniuk, line 23: Set empty string here if there are no selected elements, empty will return false if convert to boolean and there will be always string type value in your state

OleksiiBlezniuk (changes requested):

PR #1018, opened 2022-10-19
react goods

boriskerekesha:
https://boriskerekesha.github.io/react_goods-selector/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,81 @@ export const goods = [
...
-          <td>
-            <button
-              data-cy="RemoveButton"
-              type="button"
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+        {selectedGood.length
```
TarasHoliuk, line 44: use [conditional rendering](https://reactjs.org/docs/conditional-rendering.html)

src/App.tsx
```diff
@@ -15,71 +16,81 @@ export const goods = [
...
+                    onClick={() => {
+                      this.isSelectedGood(good);
+                    }}
+                    className={classNames(
+                      'button',
+                      { 'is-info': selectedGood === good },
+                    )}
+                  >
+                    {selectedGood === good ? '-' : '+'}
+                  </button>
```
TarasHoliuk, line 84: your button with `-` should remove the selected good:
image
don't forget about correct `data-cy` attribute to pass all tests

TarasHoliuk (changes requested):

src/App.tsx
```diff
@@ -15,71 +16,86 @@ export const goods = [
...
+  selectedGood: string,
+};
+export class App extends React.Component<{}, State> {
+  state = {
+    selectedGood: 'Jam',
+  };
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+  isSelectedGood = (button: HTMLButtonElement, good: string): void => {
```
Bohdan-Maksymiuk, line 27: wrong function name - it should describe what it will do

Bohdan-Maksymiuk: also you have no need to pass button - just add handler on button that you need

Bohdan-Maksymiuk: or conditionally call function what you need

vlad-demchuk (approved):

PR #1021, opened 2022-10-19
react goods selector

Arsen-Shpak:
[DEMO LINK](https://Arsen-Shpak.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,85 +1,135 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
-  'Dumplings',
-  'Carrot',
-  'Eggs',
-  'Ice cream',
-  'Apple',
-  'Bread',
-  'Fish',
-  'Honey',
-  'Jam',
-  'Garlic',
+  {
+    name: 'Dumplings',
+    id: 1,
+  },
+  {
+    name: 'Carrot',
+    id: 2,
+  },
+  {
+    name: 'Eggs',
+    id: 3,
+  },
+  {
+    name: 'Ice cream',
+    id: 4,
+  },
+  {
+    name: 'Apple',
+    id: 5,
+  },
+  {
+    name: 'Bread',
+    id: 6,
+  },
+  {
+    name: 'Fish',
+    id: 7,
+  },
+  {
+    name: 'Honey',
+    id: 8,
+  },
+  {
+    name: 'Jam',
+    id: 9,
+  },
+  {
+    name: 'Garlic',
+    id: 10,
+  },
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+interface Good {
+  name: string,
+  id:number,
+}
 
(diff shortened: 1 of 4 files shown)


ArtemHulkoDev (approved):
Everything is fine)

src/App.tsx
```diff
@@ -1,85 +1,135 @@
...
-  'Ice cream',
-  'Apple',
-  'Bread',
-  'Fish',
-  'Honey',
-  'Jam',
-  'Garlic',
+  {
+    name: 'Dumplings',
+    id: 1,
```
OleksiiBlezniuk, line 9: No need to create array of objects here, especially manually
Your products are already unique, and if you want to be sure that nothing duplicated will be added in future, you can set to key attribute string that consist from product name + it's index

src/App.tsx
```diff
@@ -1,85 +1,135 @@
...
+                <tr
+                  key={good.id}
+                  data-cy="Good"
+                  className={classNames({
+                    'has-background-success-light': goodIsSelected,
+                  })}
+                >
+                  <td>
+                    <button
+                      data-cy={goodIsSelected ? 'RemoveButton' : 'AddButton'}
```
OleksiiBlezniuk, line 109: this value also can be saved in constant, it can help you to avoid using ternary operator for button attributes values

src/App.tsx
```diff
@@ -1,85 +1,135 @@
...
+            'No goods selected'
+          ) : (
+            <>
+              {`${selectedGood.name} is selected`}
+              <button
+                aria-label="Select good"
+                data-cy="ClearButton"
+                type="button"
+                className="delete ml-3"
+                onClick={() => this.selectGood(null)}
```
OleksiiBlezniuk, line 88: let's save it to method and reuse it

src/App.tsx
OleksiiBlezniuk: Or you can create separate component for button, that will accept needed props

OleksiiBlezniuk (changes requested):
GJ, nice that you reused elements instead of duplicating them, left few comments to fix

src/App.tsx
```diff
@@ -15,71 +16,99 @@ export const goods = [
...
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+        <table className="table">
+          <tbody>
+            {goods.map((good) => {
+              const goodIsSelected = good === selectedGood;
```
Andr1yK, line 71: IMHO, if it is a bool value, you can start with "is"
```suggestion
              const isGoodSelected = good === selectedGood;
```

src/App.tsx
```diff
@@ -15,71 +16,99 @@ export const goods = [
...
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Garlic
-          </td>
-        </tr>
-      </tbody>
-    </table>
-  </main>
-);
+              return (
+                <>
```
Andr1yK, line 75: Do u need this fragment?

Andr1yK (changes requested):

Andr1yK (approved):

PR #1023, opened 2022-10-19
goods selector

korniienkokarina:
[DEMO LINK](https://korniienkokarina.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +16,88 @@ export const goods = [
...
+              >
+                <td>
+                  {good === selectedGood
+                    ? (
+                      <button
+                        data-cy="RemoveButton"
+                        className="button is-info"
+                        type="button"
+                        onClick={() => {
+                          this.setState({ selectedGood: '' });
```
vallerii, line 73: You already have function called 'clearSelectedGood', why not use it

src/App.tsx
```diff
@@ -15,71 +16,88 @@ export const goods = [
...
+                        -
+                      </button>
+                    )
+                    : (
+                      <button
+                        data-cy="AddButton"
+                        type="button"
+                        className="button"
+                        onClick={() => {
+                          this.setState({ selectedGood: good });
```
vallerii, line 85: as checklist you had to create separate methods instead of creating inline event handlers
`handleClick = () => {};

return (
  <button 
    type="button"
    onClick={handleClick}
  >
   Click me
</button>
)`

src/App.tsx
```diff
@@ -15,71 +16,92 @@ export const goods = [
...
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+  addSelectedGood = (good: string) => (
```
naziamoff, line 32: actually I don't think that this method adds something, it just sets selected good

src/App.tsx
```diff
@@ -15,71 +16,92 @@ export const goods = [
...
-              data-cy="RemoveButton"
-              type="button"
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+    return (
+      <main className="section container">
+        {!selectedGood && (<h1 className="title">No goods selected</h1>)}
```
naziamoff, line 41: - you should avoid starting with negative condition
- you use the same condition here, so you can make it a ternary 
- improve the formatting. Better to make it from the new line

src/App.tsx
```diff
@@ -15,71 +16,92 @@ export const goods = [
...
+                )}
+              >
+                <td>
+                  {good === selectedGood
+                    ? (
+                      <button
+                        data-cy="RemoveButton"
+                        className="button is-info"
+                        type="button"
+                        onClick={() => {
```
naziamoff, line 76: do you need the arrow function here?

naziamoff (changes requested):

naziamoff (approved):
🔥

PR #1025, opened 2022-10-19
made selector

poznianski:
[DEMO LINK](https://poznianski.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -18,25 +18,23 @@ export const goods = [
 
 interface State {
   selectedGood: string,
-  noMessage: string,
 }
 
 export class App extends Component<{}, State> {
   state = {
     selectedGood: 'Jam',
-    noMessage: 'No goods selected',
   };
 
   selectGood = (good: string) => {
     this.setState({ selectedGood: good });
   };
 
-  removeGood = () => {
-    this.setState({ selectedGood: '' });
+  unselectGood = () => {
+    this.setState({ selectedGood: 'No goods selected' });
   };
 
   render() {
-    const { selectedGood, noMessage } = this.state;
+    const { selectedGood } = this.state;
 
     return (
       <main className="section container">
@@ -45,7 +43,7 @@ export class App extends Component<{}, State> {
           { 'is-flex is-align-items-center': selectedGood },
         )}
         >
-          {selectedGood
+          {selectedGood !== 'No goods selected'
             ? (
               <>
                 {`${selectedGood} is selected`}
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +16,96 @@ export const goods = [
...
-    </table>
-  </main>
-);
+              return (
+                <tr
+                  data-cy="Good"
+                  key={good}
+                  className={isSelected
+                    ? 'has-background-success-light'
+                    : ''}
```
Andrew-Skoryk, line 74: `className={classNames(
   {
      'has-background-success-light': isSelected
  },
)}`

Andrew-Skoryk (approved):
GJ! You can try to use one button with checks but it's up to you=)

src/App.tsx
```diff
@@ -15,71 +16,96 @@ export const goods = [
...
+                {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+                <button
+                  data-cy="ClearButton"
+                  type="button"
+                  className="delete ml-3"
+                  onClick={this.removeGood}
+                />
+              </>
+            )
+            : 'No goods selected'}
```
naziamoff, line 60: I think we can write 'no goods selected' right inside the state, it will make your code clearer

src/App.tsx
naziamoff: @Andrew-Skoryk is right, would be better to use `classnames` library here

naziamoff (changes requested):

src/App.tsx
```diff
@@ -15,71 +16,98 @@ export const goods = [
...
+                        >
+                          -
+                        </button>
+                      )
+                      : (
+                        <button
+                          data-cy="AddButton"
+                          type="button"
+                          className="button"
+                          onClick={() => this.selectGood(good)}
```
naziamoff, line 95: do you really need to call the method with anonymous function? If you just write this.selectGood(good) it should work too

src/App.tsx
```diff
@@ -15,71 +16,98 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+interface State {
+  selectedGood: string,
+  noMessage: string,
```
naziamoff, line 21: actually you don't need separate variable in state for the message that no selected goods. I meant you could write 'no goods selected' inside selectedGood state

src/App.tsx
```diff
@@ -15,71 +16,98 @@ export const goods = [
...
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+  removeGood = () => {
```
naziamoff, line 34: wrong name for a method. Does it really remove good, or it just unselects it?

naziamoff (changes requested):

naziamoff (approved):

PR #1027, opened 2022-10-20
Goods-selector list was created

vallerii:


src/App.tsx
```diff
@@ -15,71 +16,96 @@ export const goods = [
...
-    </table>
-  </main>
-);
+              return (
+                <tr
+                  data-cy="Good"
+                  className={classNames({
+                    'has-background-success-light': goodSelected,
+                  })}
+                >
```
silchyn, line 75: We should use key prop when creating list

silchyn (changes requested):

naziamoff (approved):
GJ

PR #1028, opened 2022-10-20
Component selector has been created

MAKSOti:


dsydorenko534 (approved):
Very well😉

src/App.tsx
```diff
@@ -15,71 +16,90 @@ export const goods = [
...
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedItem: string;
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    selectedItem: 'Jam',
```
naziamoff, line 25: variable name should perfectly describe what is stored inside. The word `item` does not give me additional info

src/App.tsx
```diff
@@ -15,71 +16,90 @@ export const goods = [
...
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  saveSelectedItem = (selectedItem: string) => {
```
naziamoff, line 28: do we actually save the good with this method, or we just select it?

naziamoff: I would come up with setSelectedGood

src/App.tsx
```diff
@@ -15,71 +16,90 @@ export const goods = [
...
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+  removeSelectedItem = () => {
```
naziamoff, line 32: same here, with this name of the variable I expect that the good will be removed from the list at all. What about `unselectGood`?

src/App.tsx
```diff
@@ -15,71 +16,90 @@ export const goods = [
...
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+    return (
+      <main className="section container">
+        {selectedItem
+          ? (
+            <h1 className="title is-flex is-align-items-center">
```
naziamoff, line 43: I guess we can store `No goods selected` inside the selectedItem in the state

src/App.tsx
```diff
@@ -15,71 +16,90 @@ export const goods = [
...
+            {goods.map(item => (
+              <tr
+                data-cy="Good"
+                className={classNames({
+                  'has-background-success-light': (item === selectedGood),
+                })}
+                key={item}
+              >
+                <td>
+                  {(item === selectedGood)
```
naziamoff, line 70: unnecessary braces
```suggestion
                  { item === selectedGood
```

src/App.tsx
```diff
@@ -15,71 +16,90 @@ export const goods = [
...
+              No goods selected
+            </h1>
+          )}
+        <table className="table">
+          <tbody>
+            {goods.map(item => (
+              <tr
+                data-cy="Good"
+                className={classNames({
+                  'has-background-success-light': (item === selectedGood),
```
naziamoff, line 65: you can do it without braces

naziamoff (approved):

PR #1040, opened 2022-10-23
create react list

AndriyZhabynets:
[DEMO LINK](https://AndriyZhabynets.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import { Component } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


varkovroma (approved):

src/App.tsx
```diff
@@ -15,71 +16,95 @@ export const goods = [
...
-            Garlic
-          </td>
-        </tr>
-      </tbody>
-    </table>
-  </main>
-);
+              return (
+                <tr
+                  key={good + index.toString()}
```
sergii-nosachenko, line 67:
```suggestion
                  key={good}
```
In future don't do this - index can change for the same good, but key needs to be the same every time.

src/App.tsx
```diff
@@ -15,71 +16,95 @@ export const goods = [
...
+        {selectedGood
+          ? (
+            <h1 className="title is-flex is-align-items-center">
+              {`${selectedGood} is selected`}
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Jam
-          </td>
-        </tr>
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
sergii-nosachenko, line 47: Use `aria-label` attribute instead of disabling linter

sergii-nosachenko (approved):
Good, but read comments 👍

PR #1061, opened 2022-11-05
solution

YuriMoreau:
[DEMO LINK](https://yurimoreau.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,86 @@ export const goods = [
...
+                <td>
+                  {selectedGood === good
+                    ? (
+                      <button
+                        data-cy="RemoveButton"
+                        type="button"
+                        className="button is-info"
+                        onClick={() => {
+                          this.setState({ selectedGood: '' });
+                        }}
```
qoa11a, line 73: Move it to a function. By the way, you even already have removeSelection function. Just use it here.

src/App.tsx
```diff
@@ -15,71 +16,86 @@ export const goods = [
...
+                      </button>
+                    )
+                    : (
+                      <button
+                        data-cy="AddButton"
+                        type="button"
+                        className="button"
+                        onClick={() => {
+                          this.setState({ selectedGood: good });
+                        }}
```
qoa11a, line 85: Move it to a function.

src/App.tsx
```diff
@@ -15,71 +16,86 @@ export const goods = [
...
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  removeSelection = () => (
+    this.setState({ selectedGood: '' })
```
qoa11a, line 29: I recommend to set null as a value for the case when no good is selected.

qoa11a (changes requested):

qoa11a (approved):

PR #1069, opened 2022-11-16
solution

MelnychukYana:
https://MelnychukYana.github.io/react_goods-selector/


src/App.tsx
```diff
@@ -15,71 +16,90 @@ export const goods = [
...
+    return (
+      <main className="section container">
+        <h1 className={classNames('title',
+          { 'is-flex is-align-items-center': selectedGood })}
+        >
+          {selectedGood
+            ? (
+              <>
+                {`${selectedGood} is selected` }
+                {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
Bohdan-Maksymiuk, line 44: do not disable eslint rules - try to fix them instead

src/App.tsx
```diff
@@ -15,71 +16,90 @@ export const goods = [
...
+                <button
+                  data-cy="ClearButton"
+                  type="button"
+                  className="delete ml-3"
+                  onClick={this.removeSelection}
+                />
```
Bohdan-Maksymiuk, line 50: button in h1 tag)
looks interesting, but better conditionally prepare 1 common string and use it in h1
and conditionally render button if good is selected after it

Bohdan-Maksymiuk (changes requested):
pls, check your demo - it doesn't work
try to deploy again or check link

PR #1072, opened 2022-11-18
Goods selector task solution

juliiaap:
- [DEMO LINK](https://juliiaap.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,91 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string;
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component {
+  state: Readonly<State> = {
+    selectedGood: 'Jam',
+  };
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +15,95 @@ export const goods = [
...
-              -
-            </button>
-          </td>
+    return (
+      <main className="section container">
+        {selectedGood
+          ? (
+            <h1 className="title is-flex is-align-items-center">
+              {selectedGood}
+              {' is selected'}
```
loralevitska, line 44:
```suggestion
              {`${selectedGood} is selected`}
```

src/App.tsx
```diff
@@ -15,71 +15,95 @@ export const goods = [
...
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Jam
-          </td>
-        </tr>
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
loralevitska, line 46: try to fix this rule and don't disable linter

loralevitska (changes requested):
almost done)

src/App.tsx
juliiaap: fixed

src/App.tsx
juliiaap: fixed

polosanya (approved):
Well done!

PR #1079, opened 2022-11-21
implemented goods selector

Artem66:
[DEMO LINK](https://Artem66.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,24 +15,90 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string,
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    selectedGood: `${goods[8]} is selected`,
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  render() {
+    const { selectedGood } = this.state;
 
-    <table className="table">
-      <tbody>
-        <tr data-cy="Good">
+    return (
+      <main className="section container">
+        {selectedGood.length === 0
+          ? <h1 className="title">No goods selected</h1>
+          : (
+            <h1 className="title is-flex is-align-items-center">
+              {selectedGood}
+
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+              <button
+                data-cy="ClearButton"
+                type="button"
+                className="delete ml-3"
+                onClick={() => {
+                  this.setState({ selectedGood: '' });
+                }}
+              />
+            </h1>
+          )}
+
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
+                data-cy="Good"
+                key={good}
+                className={
+                  selectedGood === `${good} is selected`
+                    ? 'has-background-success-light'
+                    : ''
+                }
+              >
+                <td>
+                  {
+                    selectedGood !== `${good} is selected`
+                      ? (
+                        <button
+                          data-cy="AddButton"
+                          type="button"
+                          className="button"
+                          onClick={() => {
+                            this.setState({
+                              selectedGood: `${good} is selected`,
+                            });
+                          }}
+                        >
+                          +
+                        </button>
+                      )
+                      : (
+                        <button
+                          data-cy="RemoveButton"
+                          type="button"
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +15,93 @@ export const goods = [
...
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
+                data-cy="Good"
+                key={good}
+                className={
+                  selectedGood === `${good} is selected`
+                    ? 'has-background-success-light'
+                    : ''
```
Andriy-Shymkiv, line 68: better to use className library here

src/App.tsx
```diff
@@ -15,71 +15,93 @@ export const goods = [
...
-              type="button"
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+    return (
+      <main className="section container">
+        {selectedGood.length === 0
+          ? <h1 className="title">No goods selected</h1>
```
Andriy-Shymkiv, line 45: better use one h1 tag in this case because you repeat yourself here and on 47 line, you can see example in my solution

src/App.tsx
```diff
@@ -15,71 +15,93 @@ export const goods = [
...
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string,
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    selectedGood: `${goods[8]} is selected`,
```
Andriy-Shymkiv, line 24: wOw, nice solution)

Andriy-Shymkiv (changes requested):
make a little fix with h1 tag and classname library

Andriy-Shymkiv (approved):

src/App.tsx
denys-danyliuk: looks like hardcode actually, we can change position of Jam then it will break

src/App.tsx
```diff
@@ -15,71 +16,94 @@ export const goods = [
...
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  handleClickClearButton = () => {
```
denys-danyliuk, line 28: try to find a better name

src/App.tsx
```diff
@@ -15,71 +16,94 @@ export const goods = [
...
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+  handleAddButton = (currenGood: string) => {
```
denys-danyliuk, line 34: same here

src/App.tsx
```diff
@@ -15,71 +16,94 @@ export const goods = [
...
+    this.setState({
+      selectedGood: `${currenGood} is selected`,
```
denys-danyliuk, line 36:
```suggestion
      selectedGood: `${currentGood} is selected`,
```

src/App.tsx
denys-danyliuk: agree

src/App.tsx
```diff
@@ -15,71 +16,94 @@ export const goods = [
...
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
+                data-cy="Good"
+                key={good}
+                className={classNames(
+                  {
+                    'has-background-success-light':
+                      selectedGood === `${good} is selected`,
```
denys-danyliuk, line 69:
```suggestion
                    'has-background-success-light': selectedGood === good,
```

src/App.tsx
denys-danyliuk: also don't save whole string, it's called selectedGood so be it

denys-danyliuk (changes requested):

src/App.tsx
```diff
@@ -15,71 +16,93 @@ export const goods = [
...
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  clearSelectOnClick = () => {
```
denys-danyliuk, line 28:
```suggestion
  removeSelectedGood = () => {
```

src/App.tsx
```diff
@@ -15,71 +16,93 @@ export const goods = [
...
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+  selectOnClick = (currentGood: string) => {
```
denys-danyliuk, line 34:
```suggestion
  selectGood = (currentGood: string) => {
```

denys-danyliuk (approved):

PR #1084, opened 2022-11-22
solution

andrii-bassarab:
[DEMO LINK](https://andrii-bassarab.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,71 +16,103 @@ export const goods = [
...
+                  type="button"
+                  className="delete ml-3"
+                  onClick={this.handleClearGood}
+                />
+              </h1>
+            )
+            : (
+              <h1 className="title">No goods selected</h1>
+            )
+        }
```
Stepico, line 61: It looks good, but no one writes this way, I already made the same mistake, it's better not to place this curly brackets on the blank strings

src/App.tsx
```diff
@@ -15,71 +16,103 @@ export const goods = [
...
+                        onClick={() => {
+                          this.handleSelectGood(good);
+                        }}
+                      >
+                        {
+                          isSelected
+                            ? '-'
+                            : '+'
+                        }
+                      </button>
```
Stepico, line 103: It looks complex with too much logic as for me, everything is good, this way we don't repeat ourselves, but we have only 2 variations. it's not necessary to put condition into each attribute

Stepico (approved):
Seems good, just some moments I'm not sure about

Olexandr6 (changes requested):

Olexandr6 (approved):
Good job 👍

PR #1086, opened 2022-11-22
Develop

sh1gatsu:
DEMO LINK: https://sh1gatsu.github.io/react_goods-selector/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,84 @@ export const goods = [
...
-          </td>
-        </tr>
+  render() {
+    const { selectedGood } = this.state;
 
-        <tr data-cy="Good" className="has-background-success-light">
-          <td>
+    return (
+      <main className="section container">
+        <h1 className="title is-flex is-align-items-center">
```
Anna-Shapovalova, line 37: We have only class "title" when there are no selected goods

src/App.tsx
```diff
@@ -15,71 +16,84 @@ export const goods = [
...
+          {selectedGood.length
+            ? (`${selectedGood} is selected`)
+            : ('No goods selected')}
+          {selectedGood && (
```
Anna-Shapovalova, line 41: U've already checked this, so U don`t need to do it again - combine these two points

src/App.tsx
```diff
@@ -15,71 +16,84 @@ export const goods = [
...
+                      <button
+                        data-cy="RemoveButton"
+                        type="button"
+                        className="button is-info"
+                        onClick={this.resetClick}
+                      >
+                        -
+                      </button>
+                    )
+                    : (
```
Anna-Shapovalova, line 75:
```suggestion
                  {selectedGood === good ? (
                      <button
                        data-cy="RemoveButton"
                        type="button"
                        className="button is-info"
                        onClick={this.resetClick}
                      >
                        -
                      </button>
                  ) : (
```
U can put them together

src/App.tsx
```diff
@@ -15,71 +16,84 @@ export const goods = [
...
+                      <button
+                        data-cy="AddButton"
+                        type="button"
+                        className="button"
+                        onClick={() => {
+                          this.setState({ selectedGood: good });
```
Anna-Shapovalova, line 81: I guess it will be better to create a separate method as U've done with the previous one - checklist p 2.

Anna-Shapovalova (changes requested):

src/App.tsx
sh1gatsu: I like my version more ;3

src/App.tsx
```diff
@@ -15,71 +16,85 @@ export const goods = [
...
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  resetClick = () => {
```
denys-danyliuk, line 28:
```suggestion
  removeSelectedGood = () => {
```

src/App.tsx
```diff
@@ -15,71 +16,85 @@ export const goods = [
...
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
+                data-cy="Good"
+                key={good}
+                className={classNames(
+                  '',
+                  { 'has-background-success-light': selectedGood === good },
+                )}
```
denys-danyliuk, line 62:
```suggestion
                className={classNames(
                  { 
                    'has-background-success-light': selectedGood === good 
                  },
                )}
```

src/App.tsx
denys-danyliuk: it is but anyway it's better to create method for this

denys-danyliuk (approved):

PR #1094, opened 2022-11-22
Goods selector completed

Kry-Gin:
[DEMO LINK](https://kry-gin.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,89 @@ export const goods = [
...
+                data-cy="Good"
+                key={good}
+                className={classNames(
+                  {
+                    'has-background-success-light': selectedGood === good,
+                  },
+                )}
+              >
+                <td>
+                  {selectedGood !== good
```
Largoxoxo, line 71:
```suggestion
            {goods.map(good => (
              const isSelected = good === selectedGood;
              <tr
                data-cy="Good"
                key={good}
                className={classNames(
                  {
                    'has-background-success-light': isSelected,
                  },
                )}
              >
                <td>
                  {isSelected
```
Consider using a variable for 'selectedGood === good' comparison, since you use it twice in your code.

Largoxoxo (approved):
🚀

src/App.tsx
Kry-Gin: I use it only once... Oo

Largoxoxo: Yeah, my bad. I meant you use the base comparison 'selectedGood === good' and could just negate it the second time)

Olexandr6 (approved):
Good job 👍

PR #1095, opened 2022-11-22
Goods-selector done

Esceype:
- [DEMO LINK](https://Esceype.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -58,18 +58,20 @@ export class App extends Component<{}, State> {
         <table className="table">
           <tbody>
             {goods.map(good => {
+              const isSelected = good === selectedGood;
+
               return (
                 <tr
                   data-cy="Good"
                   key={good}
                   className={classNames(
                     {
-                      'has-background-success-light': good === selectedGood,
+                      'has-background-success-light': isSelected,
                     },
                   )}
                 >
                   <td>
-                    {good === selectedGood
+                    {isSelected
                       ? (
                         <button
                           data-cy="RemoveButton"


DanilSerzh (approved):
Well done, I think, I don't see any mistake

src/App.tsx
```diff
@@ -15,71 +16,90 @@ export const goods = [
...
-              className="button"
-            >
-              +
-            </button>
-          </td>
+        <table className="table">
+          <tbody>
+            {goods.map(good => {
+              return (
+                <tr
```
Olexandr6, line 62: Check the rule - [mate-academy.github.io/style-guides/javascript.html#82-if-the-function-body-consists-of-a-single-statement-returning-anexpressionwithout-side-effects-omit-the-braces-and-use-the-implicit-return-otherwise-keep-the-braces-and-use-areturnstatement](https://mate-academy.github.io/style-guides/javascript.html#82-if-the-function-body-consists-of-a-single-statement-returning-anexpressionwithout-side-effects-omit-the-braces-and-use-the-implicit-return-otherwise-keep-the-braces-and-use-areturnstatement)

src/App.tsx
```diff
@@ -15,71 +16,90 @@ export const goods = [
...
+                  data-cy="Good"
+                  key={good}
+                  className={classNames(
+                    {
+                      'has-background-success-light': good === selectedGood,
+                    },
+                  )}
+                >
+                  <td>
+                    {good === selectedGood
```
Olexandr6, line 72: Would be good to create a separate variable or function for this condition, cause you use it in several places

Olexandr6 (changes requested):

Olexandr6 (approved):

PR #1102, opened 2022-11-23
done

AndreyChernysh:
[DEMO LINK](https://AndreyChernysh.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import { Component } from 'react';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


kryvets-vadym (approved):
Good Job!

src/App.tsx
```diff
@@ -15,71 +16,87 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string | null,
```
Olexandr6, line 20: You don't need null here. Just use `''` empty string as the default value

src/App.tsx
```diff
@@ -15,71 +16,87 @@ export const goods = [
...
-              -
-            </button>
-          </td>
+    return (
+      <main className="section container">
+        {!selectedGood
+          ? <h1 className="title">No goods selected</h1>
+          : (
+            <h1 className="title is-flex is-align-items-center">
+              {`${selectedGood} is selected`}
```
Olexandr6, line 45: Let's don't use the ternary operator here. Use `||` operator instead. You don't need to duplicate `h1` twice

src/App.tsx
```diff
@@ -15,71 +16,87 @@ export const goods = [
...
+                data-cy="Good"
+                key={good}
+                className={classNames(
+                  {
+                    'has-background-success-light': good === selectedGood,
+                  },
+                )}
+              >
+                <td>
+                  {good === selectedGood
```
Olexandr6, line 70: You use this codition in several places, so makes sense to move it to a variable or some function, like we did today on the lesson

Olexandr6 (changes requested):

src/App.tsx
Olexandr6: you didn't fix this comment

Olexandr6 (changes requested):

src/App.tsx
```diff
@@ -15,71 +16,101 @@ export const goods = [
...
+    this.setState({ selectedGood: good });
+  };
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+  render() {
+    const { selectedGood } = this.state;
+    const isGoodSelected = selectedGood.length === 0;
```
Olexandr6, line 38:
```suggestion
    const isGoodsSelected = selectedGood.length === 0;
```

Olexandr6 (approved):

PR #1103, opened 2022-11-23
table

sofiia-dovhan:
[DEMO LINK](https://sofiia-dovhan.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +15,92 @@ export const goods = [
...
-            >
-              +
-            </button>
-          </td>
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+              <button
+                data-cy="ClearButton"
+                type="button"
+                className="delete ml-3"
+                onClick={() => this.setState({ selectedGood: '' })}
```
Andriy-Shymkiv, line 55: you should create a method for this case and use it here and on 76 line.

src/App.tsx
```diff
@@ -15,71 +15,92 @@ export const goods = [
...
+                  : ''}
+              >
+                {this.isSelected(good)
+                  ? (
+                    <td>
+                      <button
+                        data-cy="RemoveButton"
+                        type="button"
+                        className="button is-info"
+                        onClick={() => this.setState({ selectedGood: '' })}
```
Andriy-Shymkiv, line 76: see above

src/App.tsx
```diff
@@ -15,71 +15,92 @@ export const goods = [
...
+    this.setState({ selectedGood: good });
+  };
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+  componentDidMount = () => {
+    this.setState({ selectedGood: 'Jam' });
+  };
```
Andriy-Shymkiv, line 37: delete this method because you don't need him, and just set `'Jam'` as a default value in state

Andriy-Shymkiv (changes requested):
you should change your code a bit

src/App.tsx
```diff
@@ -15,71 +16,95 @@ export const goods = [
...
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  isSelected = (good: string) => {
+    return good === this.state.selectedGood;
+  };
```
Olexandr6, line 30: Check the rule - https://mate-academy.github.io/style-guides/javascript.html#82-if-the-function-body-consists-of-a-single-statement-returning-anexpressionwithout-side-effects-omit-the-braces-and-use-the-implicit-return-otherwise-keep-the-braces-and-use-areturnstatement

Olexandr6 (changes requested):
One small thing is left to fix

Olexandr6 (approved):

PR #1104, opened 2022-11-23
solution

mitattt:
[DEMO LINK](https://mitattt.github.io/react_goods-selector/)

For some reason demo link does not work after deploy

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +16,84 @@ export const goods = [
...
-              +
-            </button>
-          </td>
+    return (
+      <main className="section container">
+        <h1 className={classNames('title', {
+          'is-flex is-align-items-center': selectedGood,
+        })}
+        >
+          {selectedGood
```
Largoxoxo, line 37: https://mate-academy.github.io/style-guides/javascript.html#153use-shortcuts-for-booleans-but-explicit-comparisons-for-strings-and-numbers

vadymbaranov (approved):
Looks great!

src/App.tsx
Olexandr6: Let's fix that

src/App.tsx
```diff
@@ -15,71 +16,84 @@ export const goods = [
...
-            Jam
-          </td>
-        </tr>
+                        this.setState({ selectedGood: value });
+                      }}
+                    >
+                      {isSelected
+                        ? '-'
+                        : '+'}
+                    </button>
```
Olexandr6, line 84: Would be better to have 2 buttons instead. Yesterday in the lesson, we've discussed this point. Let's fix that

G-YaSha (approved):

G-YaSha (changes requested):
.

G-YaSha (approved):
Good job 👍

PR #1108, opened 2022-11-23
Goods selector created

vadymbaranov:
[DEMO LINK](https://Largoxox.github.io/react_goods-selector/)

I deployed, but demo link still won't work((

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+// import { event } from 'cypress/types/jquery';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,92 @@ export const goods = [
...
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+              <button
+                data-cy="ClearButton"
+                type="button"
+                className="delete ml-3"
+                onClick={this.clearSelectedGoodOnClick}
+              />
+            </h1>
+          ) : (
+            <h1 className="title">No goods selected</h1>
```
mitattt, line 54: u could use classNames here to avoid double calling tag h1

mitattt (approved):
great job 🚀

src/App.tsx
Largoxoxo: Yeah, I tried but it didn't function properly, so I went back to a more straightforward option. Will try later. Thanks for the tip though)

G-YaSha (approved):
Good job 👍

PR #1118, opened 2022-11-28
goods selector

Bohdan-Kovalenko1992:
[DEMO LINK](https://Bohdan-Kovalenko1992.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,82 @@ export const goods = [
...
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  handleClearButton = () => {
```
masha-girya, line 28: here you have two verbs one by one in method name, you should leave only one verb: clearSelection / removeSelection / removeGood / clearGood

src/App.tsx
```diff
@@ -15,71 +16,82 @@ export const goods = [
...
-            Garlic
-          </td>
-        </tr>
-      </tbody>
-    </table>
-  </main>
-);
+              className="delete ml-3"
+              onClick={this.handleClearButton}
+            />
```
masha-girya, line 55: you can add button right to the 45 row where you have condition when good is selected

src/App.tsx
```diff
@@ -15,71 +16,82 @@ export const goods = [
...
+              <tr
+                key={good}
+                data-cy="Good"
+                className={classNames(
+                  { 'has-background-success-light': selectedGood === good },
+                )}
+              >
+                <td>
+                  <button
+                    data-cy={selectedGood === good
```
masha-girya, line 70: you have selectedGood === good in 65, 70, 76, 82 rows, it will be better to move this condition into variable

src/App.tsx
```diff
@@ -15,71 +16,82 @@ export const goods = [
...
+                      { 'is-info': selectedGood === good },
+                    )}
+                    onClick={() => {
+                      this.handleAddRemoveButton(good);
+                    }}
+                  >
+                    {selectedGood === good
+                      ? '-'
+                      : '+'}
+                  </button>
```
masha-girya, line 85: also, it is hard to read(( It is better to make two separate button and show one of them depending in condition

masha-girya (changes requested):
nice job! You have something to improve) also, Demo link doesn't work

PR #1124, opened 2022-12-04
Solution

berko228:
[DEMO LINK](https://berko228.github.io/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +16,85 @@ export const goods = [
...
-        </tr>
+        <table className="table">
+          <tbody>
+            {goods.map((good) => (
+              <tr
+                data-cy="Good"
+                key={good}
+                className={selectedGood === good
+                  ? 'has-background-success-light'
+                  : ''}
```
nazarmatsevych, line 62: Use classNames library for adding class conditionally

src/App.tsx
```diff
@@ -15,71 +16,85 @@ export const goods = [
...
+                    type="button"
+                    className={classNames(
+                      'button',
+                      {
+                        'button is-info': selectedGood === good,
+                      },
+                    )}
+                    onClick={selectedGood === good
+                      ? this.removeSelected
+                      : () => {
```
TarasHoliuk, line 78: be consistent - create a separate function for selecting good as well

nazarmatsevych (changes requested):
Good work, check comment

src/App.tsx
```diff
@@ -15,71 +16,85 @@ export const goods = [
...
+                <td>
+                  <button
+                    data-cy={selectedGood === good
+                      ? 'RemoveButton'
+                      : 'AddButton'}
+                    type="button"
+                    className={classNames(
+                      'button',
+                      {
+                        'button is-info': selectedGood === good,
```
TarasHoliuk, line 73:
```suggestion
                        'is-info': selectedGood === good,
```
`button` classname always exists - 71 row

src/App.tsx
```diff
@@ -15,71 +16,85 @@ export const goods = [
...
-            Jam
-          </td>
-        </tr>
+        <table className="table">
+          <tbody>
+            {goods.map((good) => (
+              <tr
+                data-cy="Good"
+                key={good}
+                className={selectedGood === good
```
TarasHoliuk, line 60: use classnames lib

TarasHoliuk (changes requested):

src/App.tsx
tetiana-ivanchenko: and this
image

tetiana-ivanchenko (changes requested):
when you re-request changes, make sure that all prev comments are fixed

src/App.tsx
```diff
@@ -15,71 +16,85 @@ export const goods = [
...
+                        'is-info': selectedGood === good,
+                      },
+                    )}
+                    onClick={selectedGood === good
+                      ? this.removeSelected
+                      : () => {
+                        return (
+                          this.setState({ selectedGood: good })
+                        );
+                      }}
```
OlEzhA0, line 82: you can create some function and decide there what do you need to do (remove or select)
try to don't put conditional operation to events

src/App.tsx
```diff
@@ -15,71 +16,85 @@ export const goods = [
...
+                  >
+                    {selectedGood === good
+                      ? '-'
+                      : '+'}
```
OlEzhA0, line 86:
```suggestion
                    {selectedGood === good ? '-' : '+'}
```

src/App.tsx
```diff
@@ -15,71 +16,85 @@ export const goods = [
...
+                    data-cy={selectedGood === good
+                      ? 'RemoveButton'
+                      : 'AddButton'}
+                    type="button"
+                    className={classNames(
+                      'button',
+                      {
+                        'is-info': selectedGood === good,
+                      },
+                    )}
```
OlEzhA0, line 75:
```suggestion
                    className={classNames(
                      'button', 
                      { 'is-info': selectedGood === good },
                    )}
```

OlEzhA0, line 75: move `selectedGood === good` to variable and use it where needed

OlEzhA0 (changes requested):
Hey! Good job
I left some recommendations, please check it
and resolve issues from other mentors

src/App.tsx
berko228: there is no point of moving it to variable, just because i'll do it like a function, the function will have two params and as result, the code will be even longer than now. I had a conversation in fe_chat2 and mentor said taht i'm right

tetiana-ivanchenko (approved):

src/App.tsx
OlEzhA0: Hey! thank you for your point
I suggested to do something like 

```
{goods.map(good => {
  const isSelectedGood = selectedGood === good
  
  return <>JSX</>
})}
```

It's like DRY principal, for example you used this condition in 10 places and in whole project (it's can be) and you managers changed requirements and you need to fix `selectedGood === good  to selectedGood !== good `

It's easier to implement in 1 place and use this constant wherever than find the condition and change in whole project 

In programming we haven’t goal to have short or long code. Code should be readable and easily maintainable

PR #1127, opened 2022-12-07
goods selector

Mikitkaaist:


src/App.tsx
```diff
@@ -15,71 +16,75 @@ export const goods = [
...
+                  className={classNames({
+                    'has-background-success-light': isSelected,
+                  })}
+                >
+                  <td>
+                    <button
+                      type="button"
+                      onClick={() => (isSelected
+                        ? this.setState({ selectedGood: '' })
+                        : this.setState({ selectedGood: good }))}
```
tetiana-ivanchenko, line 67:
```suggestion
                      onClick={() => this.setState({ 
                        selectedGood: isSelected ? '' : good
                      }))}
```

tetiana-ivanchenko: actually, if you have some kind of condition, it means that you need a separate function for it

tetiana-ivanchenko (changes requested):

src/App.tsx
```diff
@@ -15,71 +16,75 @@ export const goods = [
...
+                  className={classNames({
+                    'has-background-success-light': isSelected,
+                  })}
+                >
+                  <td>
+                    <button
+                      type="button"
+                      onClick={() => this.setState({
+                        selectedGood: isSelected ? '' : good,
+                      })}
```
AlisaVlasova, line 67: wright separate handler for it
there you can check if incoming good is already selected or not

AlisaVlasova (changes requested):

nazarmatsevych (approved):

PR #1130, opened 2022-12-13
Solution

UmizDemud:
[DEMO LINK](https://UmizDemud.github.io/react_goods-selector/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,3 +1,7 @@
 iframe {
   display: none;
 }
+
+.button {
+  min-width: 43px;
+}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,88 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  allGoods: string[];
+  selectedGood: string;
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state = {
+    allGoods: goods,
+    selectedGood: 'Jam',
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  handleClearButtonClick = () => {
+    this.setState({ selectedGood: '' });
+  };
 
-    <table className="table">
-      <tbody>
-        <tr data-cy="Good">
-          <td>
+  render() {
+    const {
+      allGoods,
+      selectedGood,
+    } = this.state;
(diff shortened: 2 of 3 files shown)


src/App.tsx
```diff
@@ -15,71 +15,88 @@ export const goods = [
...
+        <table className="table">
+          <tbody>
+            {allGoods.map(good => (
+              <tr
+                key={good}
+                data-cy="Good"
+                className={`${selectedGood === good
+                  ? 'has-background-success-light'
+                  : ''
+                }`}
```
naziamoff, line 67: the time has come to get acquainted with the great [`classnames`](https://www.google.com/search?q=classnames&oq=classnames&aqs=chrome..69i57j0i512l6j69i60.1530j0j7&sourceid=chrome&ie=UTF-8) library)

src/App.scss
```diff
@@ -1,3 +1,7 @@
 iframe {
   display: none;
 }
+
+.button {
+  min-width: 43px;
```
naziamoff, line 6: seems like a magic number to me.
Usually the props like width, height, paddings, margins etc. should be theoretically divided by 4 without remainder

naziamoff (changes requested):
looks really good)
But let's learn new library here

src/App.tsx
UmizDemud: I have seen it being used. Wanted to try my implementation. For sure there would be mistakes. I'd be very happy to hear some obvious mistakes or comments.

src/App.scss
naziamoff: will do for the training project 👍 
Will be much easier with sass on the real one

src/scripts/classname.ts
```diff
@@ -0,0 +1,19 @@
+type Rule = {
+  [key: string]: boolean;
+};
+
+type Input = Array<string | Rule>;
+
+export const classname = (...rules: Input): string => (
```
naziamoff, line 7: impressive job man 🔥 
But I don't see you have the constant classes covered here. I mean, classnames should also have the possibility to add styles that don't need condition. We can do it like: classnames
```
cn('some-class ml-8 another-class', {
   'has-background-success-light': selectedGood === good 
})
```
where in this case the classnames `some-class ml-8 another-class` will be applied anyway, and the class 'has-background-success-light' only if the condition is true

naziamoff (approved):
Well done 👍 🔥  
There's one comment about your custom classnames script, but definitely I can't reject that. Keep going, hit on re-review if you implement the constant classes and want me to take a look again

src/scripts/classname.ts
UmizDemud: I actually have it here 😊 if a rule is only a string just pastes it to the list.
That's why the input can be string too.

```ts
    if (typeof rule === 'string') {
      return `${classes} ${rule}`;
    }
 ```
 
thanks a lot, happy to read these

PR #1137, opened 2022-12-17
Solution

Julia2063:
[DEMO LINK](https://julia2063.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,95 @@ export const goods = [
...
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string,
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
```
tetiana-ivanchenko, line 23: why do you reqrite component from functional to class?

tetiana-ivanchenko (changes requested):

src/App.tsx
Julia2063: The main difference between a class component and functionality is the ability to save state. How can i change state of component without state?

SanyaBratashchuk (approved):

PR #1153, opened 2022-12-29
solution

lefteris-L:
[DEMO LINK](https://lefteris-l.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,12 @@
-import React from 'react';
+/* eslint-disable jsx-a11y/control-has-associated-label */
+
+import { Component, FC, ReactNode } from 'react';
+import cn from 'classnames';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
 
-export const goods = [
+const goods = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,71 +19,98 @@ export const goods = [
...
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+              return (
+                <tr
+                  data-cy="Good"
```
nurdanokcu, line 79: I couldn't find a key added to the `<tr>...</tr>`

nurdanokcu (approved):
Amazing work :100:

src/App.tsx
lefteris-L: My bad

src/App.tsx
```diff
@@ -15,71 +19,99 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = { selectedGood: string };
```
naziamoff, line 22:
```suggestion
type State = { 
  selectedGood: string 
};
```

src/App.tsx
```diff
@@ -15,71 +19,99 @@ export const goods = [
...
+
+export class App extends Component<{}, State> {
+  state = { selectedGood: 'Jam' };
+
+  isSelected = (item: string) => (
```
naziamoff, line 27: from its name it should be clear that it is a function
```suggestion
  checkIsSelected = (item: string) => (
```

naziamoff (approved):
great job! A couple of suggestions

PR #1182, opened 2023-01-03
Add solution

VladimirEdelmann:
# Task
You are given an array of goods. Render them in a table with the ability to select one or clear selection.

1. Write everything inside the `App` (**don't** create additional components).
2. Save a `selectedGood` in the state (`Jam` is the default value).
3. Show the name of the selected good in the `h1.title` (`Jam is selected`).
4. Add the `has-background-success-light` class to the `tr` of the selected Good.
5. Show the `ClearButton` button in the title only when a good is selected.
6. `ClearButton` should clear selection by setting an empty string to `selectedGood`.
7. When there is no selected good, the title should show `No goods selected`.
8. Each good should have an `AddButton` to select the good.
    - only 1 good can be selected at a time;
9. Don't show `AddButton` when a good is selected.
10. Show `RemoveButton` for the selected good to clear selection.

# Implementation
[DEMO LINK](https://vladimiredelmann.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


kreisberg (approved):
Good!

src/App.tsx
```diff
@@ -15,71 +15,107 @@ export const goods = [
...
+    this.setState({
+      selectedGood: '',
+    });
+  };
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+  clearGood = () => {
```
solaryasha, line 39: What is the difference between `removeGood` and `clearGood` methods ?

src/App.tsx
```diff
@@ -15,71 +15,107 @@ export const goods = [
...
+          <tbody>
+            {
+              goods.map(good => {
+                const isGoodSelected = this.state.selectedGood === good || '';
+
+                return (
+                  <tr
+                    data-cy="Good"
+                    className={
+                      isGoodSelected && 'has-background-success-light'
```
solaryasha, line 77: let's try the `classnames` lib

solaryasha (changes requested):

solaryasha (approved):
👍

PR #1205, opened 2023-01-04
Develop

sercansimsek:


Rijo-J (approved):
Looks super legit 🚀 👍

src/App.tsx
```diff
@@ -15,71 +17,85 @@ export const goods = [
...
+  render() {
+    const { selectedGood } = this.state;
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+    return (
+      <main className="section container">
+        {selectedGood ? (
```
naziamoff, line 42:
```suggestion
        {selectedGood 
          ? (
```

src/App.tsx
```diff
@@ -15,71 +17,85 @@ export const goods = [
...
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+              className="delete ml-3"
+            />
+          </h1>
+        ) : (
+          <h1 className="title">No goods selected</h1>
```
naziamoff, line 54: actually it would be better to conditionally render inside the h1 tag and assign classnames with help of the `classnames` lib

src/App.tsx
```diff
@@ -15,71 +17,85 @@ export const goods = [
...
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+  addButtonHandler = (good: string) => {
```
naziamoff, line 33: from its name to me it seems that we add some button handler, which I believe to be some event listener or sth like that.
Basically we can call it just `handleAdd`. This looks much simpler

naziamoff (approved):
well done. A couple of suggestions though

PR #1209, opened 2023-01-05
solution

Lun4trick:
[DEMO LINK](https://lun4trick.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


lukacsja (approved):

src/App.tsx
```diff
@@ -15,71 +16,88 @@ export const goods = [
...
+type State = {
+  selectedGood: string,
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state = {
+    selectedGood: 'Jam is selected',
+  };
```
naziamoff, line 26: `selectedGood: Jam is selected` looks really wrong. From the the name of the variable I expect that it's gonna be just `Jam`, but not the whole string to render. Let's keep only the good here

src/App.tsx
```diff
@@ -15,71 +16,88 @@ export const goods = [
...
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  handleClickAdd = (item: string) => (
+    this.state.selectedGood === `${item} is selected`
+      ? this.setState({ selectedGood: 'No goods selected' })
+      : this.setState({ selectedGood: `${item} is selected` })
```
naziamoff, line 31: looks kinda hardcoded. Checking for exact match with a strings is not the best option. You'll handle it better if you fix my previous comment

src/App.tsx
```diff
@@ -15,71 +16,88 @@ export const goods = [
...
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+  render() {
+    const { selectedGood } = this.state;
+    const checkItemIsSelected = (selectedGood !== 'No goods selected');
```
naziamoff, line 36: same here

src/App.tsx
```diff
@@ -15,71 +16,88 @@ export const goods = [
...
+        </h1>
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Jam
-          </td>
-        </tr>
+        <table className="table">
+          <tbody>
+            {goods.map(item => {
+              const itemcheck = selectedGood === `${item} is selected`;
```
naziamoff, line 60: booleans should be named as booleans, starting with `should`, `can`, `is` and so on

naziamoff (changes requested):

naziamoff (changes requested):
you did not fix all of my previous comments

src/App.tsx
```diff
@@ -15,71 +16,90 @@ export const goods = [
...
-          </td>
-        </tr>
+        <h1 className="title is-flex is-align-items-center">
+          {isSelected
+            ? `${selectedGood} is selected`
+            : 'No goods selected'}
 
-        <tr data-cy="Good" className="has-background-success-light">
-          <td>
+          { (isSelected) && (
```
naziamoff, line 46:
```suggestion
          {isSelected && (
```

src/App.tsx
```diff
@@ -15,71 +16,90 @@ export const goods = [
...
+        </h1>
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Jam
-          </td>
-        </tr>
+        <table className="table">
+          <tbody>
+            {goods.map(item => {
+              const itemcheck = selectedGood === item;
```
naziamoff, line 62: that's what I meant, your boolean variables should be named as booleans. Here you have `itemcheck`

src/App.tsx
```diff
@@ -15,71 +16,90 @@ export const goods = [
...
+                  key={item}
+                  data-cy="Good"
+                  className={classNames(
+                    { 'has-background-success-light': itemcheck },
+                  )}
+                >
+                  <td>
+                    <button
+                      data-cy={
+                        (itemcheck)
```
naziamoff, line 75:
```suggestion
                        itemcheck
```

naziamoff (approved):
I approve, but please consider those in your next tasks

PR #1219, opened 2023-01-05
Finished

JzSol:
[DEMO LINK](https://JzSol.github.io/react_goods-selector/)


naziamoff (approved):
not bad in general, but there is some issues with the code style

PR #1223, opened 2023-01-06
solar system

dolnaro:
[DEMO LINK](https://dolnaro.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,88 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string;
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state = {
+    selectedGood: 'Jam',
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +15,88 @@ export const goods = [
...
+    this.setState({
+      selectedGood: '',
+    });
+  };
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+  renderGood = (good: string) => {
```
Grayfox147, line 39: Interesting approach, separating and creating a helping method.

src/App.tsx
```diff
@@ -15,71 +15,88 @@ export const goods = [
...
-        </tr>
-
-        <tr data-cy="Good">
-          <td>
+          ) : (
             <button
               data-cy="AddButton"
               type="button"
               className="button"
+              onClick={() => this.selectGood(good)}
```
Grayfox147, line 63: also quite cool to save the comparison or the checking of the selected good into a variable!

Grayfox147 (commented):
I believe it is quite different than my approach but you reach to the same result!, really particular and smart approach!

src/App.tsx
```diff
@@ -15,71 +15,88 @@ export const goods = [
...
-    </table>
-  </main>
-);
+  render() {
+    return (
+      <main className="section container">
+        {this.state.selectedGood ? (
+          <h1 className="title is-flex is-align-items-center">
+            {this.state.selectedGood}
+            {' is selected'}
```
naziamoff, line 82: why not just use interpolation here?

naziamoff (approved):
well done! Just one small thing

PR #1225, opened 2023-01-07
solution

nzrii-kysel:
[DEMO LINK](https://nzrii-kysel.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


Malva37 (approved):
Good job)

src/App.tsx
```diff
@@ -15,71 +15,80 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selected: string,
```
Anna-Tkachenko, line 19: will be better to specify some noun in variable name, for now it's not clear what entity you want to store in state

src/App.tsx
```diff
@@ -15,71 +15,80 @@ export const goods = [
...
-          </td>
-        </tr>
-      </tbody>
-    </table>
-  </main>
-);
+              return (
+                <tr
+                  data-cy="Good"
+                  className={isSelected ? 'has-background-success-light' : ''}
```
Anna-Tkachenko, line 59: use classnames library if you need to set up classname according to some condition

src/App.tsx
```diff
@@ -15,71 +15,80 @@ export const goods = [
...
+                          onClick={this.clearSelected}
+                        >
+                          -
+                        </button>
+                      ) : (
+                        <button
+                          data-cy="AddButton"
+                          type="button"
+                          className="button"
+                          onClick={() => this.setState({ selected: good })}
```
Anna-Tkachenko, line 77: check point 2 in checklist

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #1232, opened 2023-01-20
Goods_selector

katya-suvorova:
https://katya-suvorova.github.io/react_goods-selector/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,84 @@ export const goods = [
...
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  handleButtonClear = () => {
```
agniya-i, line 28:
```suggestion
  handleСlear = () => {
```
it would be better to change to `handleClear`, `handleButtonClear` - seems we want to clear the button

agniya-i (approved):

PR #1236, opened 2023-01-23
Solution

RomeoKs:
[DEMO LINK](https://RomeoKs.github.io/react_goods-selector/)


src/App.tsx
```diff
@@ -15,71 +15,80 @@ export const goods = [
...
+export const App: React.FC = () => {
+  const [selectedGood, setSelectedGood] = useState('Jam');
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+  const goodsList = goods.map(good => (
+    <tr
+      data-cy="Good"
+      key={good}
+      className={
```
naziamoff, line 25: [classnames](https://www.npmjs.com/package/classnames)

src/App.tsx
```diff
@@ -15,71 +15,80 @@ export const goods = [
...
+        (selectedGood === good)
+          ? ('has-background-success-light')
+          : ''
+      }
+    >
+      <td>
+        {
+          (selectedGood !== good)
+            ? (
```
naziamoff, line 34:
```suggestion
        {selectedGood !== good
            ? (
```

naziamoff: also, it is better to start with positive condition

src/App.tsx
```diff
@@ -15,71 +15,80 @@ export const goods = [
...
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Jam
-          </td>
-        </tr>
+  return (
+    <main className="section container">
+      <h1 className="title is-flex is-align-items-center">
+        {
+          (!selectedGood
+            ? 'No goods selected'
```
naziamoff, line 69:
```suggestion
        {!selectedGood
          ? 'No goods selected'
```

naziamoff: start with positive.
The point is that negative conditions are more difficult to compile in one's head. I read it like 
`No selected goods ? yes : no`. It takes some time to understand what is `no selected goods - yes`, cause as a human I mostly think in human categories, not as `true/false`

src/App.tsx
```diff
@@ -15,71 +15,80 @@ export const goods = [
...
+            : `${selectedGood} is selected`
+          )
+        }
 
-        <tr data-cy="Good">
-          <td>
+        {
+          selectedGood && (
```
naziamoff, line 75:
```suggestion
        {selectedGood && (
```

naziamoff (changes requested):

src/App.tsx
RomeoKs: Done.
Thank you!

src/App.tsx
RomeoKs: Doesn't matter as the code written to be easily to read?
For me personally it's easier to scroll the code and find it, if I have it on separate lines, than to be "glued" like this.

Anyway, I made the changes as requested.

Thank you!

naziamoff: to my mind this way we do not lose any clearance of the code, but it looks more ergonomical

PR #1239, opened 2023-01-24
NVyrsta

NVyrsta:
[DEMO LINK](https://NVyrsta.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import { Component } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


Malva37 (approved):
Optimized way of solution, great

src/App.tsx
```diff
@@ -15,71 +16,82 @@ export const goods = [
...
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+        <table className="table">
```
alexandra-protyanova, line 53: as for me, goods looks like a list, not a table) Nice practice with tables, I think we should leave it)

alexandra-protyanova (approved):
Good job 👏

PR #1246, opened 2023-01-25
solution

igor-kashchenko:
[DEMO LINK](https://igor-kashchenko.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -38,38 +38,37 @@ export class App extends Component<{}, State> {
 
     return (
       <main className="section container">
-        {selectedGood
-          ? (
-            <h1 className="title is-flex is-align-items-center">
-              {selectedGood}
+        {selectedGood ? (
+          <h1 className="title is-flex is-align-items-center">
+            {`${selectedGood} is selected`}
 
-              {' '}
-
-              is selected
-
-              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-              <button
-                data-cy="ClearButton"
-                type="button"
-                className="delete ml-3"
-                onClick={this.deleteSelect}
-              />
-            </h1>
-          )
-          : (<h1 className="title">No goods selected</h1>)}
+            <button
+              data-cy="ClearButton"
+              aria-label="Clear"
+              type="button"
+              className="delete ml-3"
+              onClick={this.deleteSelect}
+            />
+          </h1>
+        ) : (
+          <h1 className="title">No goods selected</h1>
+        )}
 
         <table className="table">
           <tbody>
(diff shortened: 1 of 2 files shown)


OlhaYevstifieieva (approved):
Nice job)

src/App.tsx
```diff
@@ -15,71 +16,89 @@ export const goods = [
...
-            >
-              -
-            </button>
-          </td>
+    return (
+      <main className="section container">
+        {selectedGood
+          ? (
+            <h1 className="title is-flex is-align-items-center">
+              {selectedGood}
```
Anna-Tkachenko, line 44: you can also write it in 1 row instead of 3
```suggestion
              {`${selectedGood} is selected`}
```

src/App.tsx
```diff
@@ -15,71 +16,89 @@ export const goods = [
...
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Garlic
-          </td>
-        </tr>
-      </tbody>
-    </table>
-  </main>
-);
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
Anna-Tkachenko, line 50: will be better to find solution instead of disabling linter

src/App.tsx
```diff
@@ -15,71 +16,89 @@ export const goods = [
...
+              <button
+                data-cy="ClearButton"
+                type="button"
+                className="delete ml-3"
+                onClick={this.deleteSelect}
+              />
+            </h1>
+          )
+          : (<h1 className="title">No goods selected</h1>)}
```
Anna-Tkachenko, line 59:
```suggestion
          : (
            <h1 className="title">No goods selected</h1>
          )}
```

src/App.tsx
```diff
@@ -15,71 +16,89 @@ export const goods = [
...
+
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
+                data-cy="Good"
+                key={good}
+                className={classNames('',
```
Anna-Tkachenko, line 67: empty string as class name is redundant here

src/App.tsx
```diff
@@ -15,71 +16,89 @@ export const goods = [
...
+                  { 'has-background-success-light': good === selectedGood })}
```
Anna-Tkachenko, line 68: you check is current good selected several times
let's create separate function or boolean variable that will store or calculate it one time, and will use it where you need it

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #1254, opened 2023-01-29
add solution

Anastasiia-Khudych:
[Demo](https://anastasiia-khudych.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,104 @@ export const goods = [
...
+    if (condition) {
+      this.setState({
+        selectedGood: '',
+      });
+    } else {
+      this.setState({
+        selectedGood: nameToSet,
+      });
+    }
+  };
```
tania-boiar, line 38: Interesting solution!

tania-boiar (approved):
Good job! 🔥

src/App.tsx
Anna-Tkachenko: to make this code clearer, you can pass needed value directly to this function (empty string or good string)
or will be better to create the variable at the start of the function, that will be assigned to the needed value according to some condition, and then call once setState
because for now you repeated some code twice

Anna-Tkachenko (approved):
The less lines of code - it's not always the best solution
Will be better to pay attention to code readability, to make it the more understandable and clean, as you can 🧚‍♀️

PR #1260, opened 2023-01-30
Solution

maasiimka:


src/App.tsx
```diff
@@ -15,71 +19,96 @@ export const goods = [
...
+                  data-cy="ClearButton"
+                  type="button"
+                  className="delete ml-3"
+                  onClick={clearGood}
+                >
+                  {}
+                </button>
+              </div>
+            )
+        }
```
vanst1, line 57:
```suggestion
        {
          selectedGood
            ? (
              <h1 className="title is-flex is-align-items-center">
                {`${selectedGood} is selected`}
                {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
                <button
                  data-cy="ClearButton"
                  type="button"
                  className="delete ml-3"
                  onClick={clearGood}
                />
              </h1>
            )
            : <h1 className="title">No goods selected</h1>
        }
```
Fix bug from video:)

https://user-images.githubusercontent.com/119679758/215580679-2b192b87-8819-4d62-a29c-48c5dbe0c0cd.MOV

vanst1 (changes requested):
Good job)
Some fix needed))

src/App.tsx
```diff
@@ -15,71 +19,93 @@ export const goods = [
...
-        </tr>
+        <table className="table">
+          <tbody>
+            {
+              goods.map((good) => (
+                selectedGood === good
+                  ? (
+                    <tr
+                      data-cy="Good"
+                      className="has-background-success-light"
```
Anna-Tkachenko, line 66: use classNames library for setting needed class according to some check
you don't need to describe `td` row 2 times, if only classNames property is differs for them

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -15,71 +20,91 @@ export const goods = [
...
+                  className={
+                    classNames({
+                      'has-background-success-light': selectedGood === good,
+                    })
+                  }
+                  key={good}
+                >
+                  <td>
+                    {
+                      selectedGood === good
```
Anna-Tkachenko, line 74: you checked is current good selected several times
create separate function for this purpose or create boolean variable with this value

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #1261, opened 2023-01-30
React Goods Selector

nkondd:
[DEMO LINK](https://nkondd.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,91 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string,
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state = {
+    selectedGood: 'Jam',
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  clearSelectedGood = () => {
+    this.setState({
+      selectedGood: '',
+    });
+  };
 
-    <table className="table">
-      <tbody>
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +15,91 @@ export const goods = [
...
-            </button>
-          </td>
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
+                data-cy="Good"
+                className={good === selectedGood
+                  ? 'has-background-success-light'
+                  : ''}
```
ihor-soloviov, line 68: `
                className={classNames({
                  'has-background-success-light': selectedGood === good,
                })}
              >
`

ihor-soloviov: I don't know how to put my version of solution normally, hahah
I think you don't need ternary operation here

ihor-soloviov (commented):
Extra good job! That was funny that you're checking for NOT something in ternary usually, haha.

src/App.tsx
```diff
@@ -15,71 +15,91 @@ export const goods = [
...
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+    return (
+      <main className="section container">
+        {!selectedGood
+          ? (
+            <h1 className="title">No goods selected</h1>
```
Anna-Tkachenko, line 46:
```suggestion
            <h1 className="title">
              No goods selected
            </h1>
```

src/App.tsx
Anna-Tkachenko: use classnames library if you need to add class according to some check

src/App.tsx
```diff
@@ -15,71 +15,91 @@ export const goods = [
...
+          <tbody>
+            {goods.map(good => (
+              <tr
+                data-cy="Good"
+                className={good === selectedGood
+                  ? 'has-background-success-light'
+                  : ''}
+              >
+                <td>
+                  {good !== selectedGood
```
Anna-Tkachenko, line 71: you checked is current good selected several times, create separate boolean variable or separate function for this purpose

denys-cheporniuk (approved):

PR #1263, opened 2023-01-30
Develop

Oleksandr-Maly:
[DEMO LINK](https://oleksandr-maly.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -25,6 +25,10 @@ export class App extends React.Component<{}, State> {
     selectedGood: 'Jam',
   };
 
+  clearGood = () => {
+    this.setState({ selectedGood: '' });
+  };
+
   render() {
     const { selectedGood } = this.state;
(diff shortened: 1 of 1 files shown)


vanst1 (approved):
Well done)

src/App.tsx
```diff
@@ -15,71 +16,93 @@ export const goods = [
...
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+            return (
+              <tbody key={good}>
```
Anna-Tkachenko, line 58: you don't need to render `tbody` for each good
`tbody` should be only one for the whole table

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job! ⭐️

PR #1264, opened 2023-01-30
realised solution to react_goods

andrii-kurash:
[DEMO LINK](https://andrii-kurash.github.io/react_goods-selector/)

Have a great mood!🔥

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,86 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+export class App extends React.Component {
+  state = {
+    selectedGood: 'Jam',
+  };
```
andrii-andriushchenko, line 22: state doesn't have typization (type / interface), pls add it before `class App`

src/App.tsx
```diff
@@ -15,71 +16,86 @@ export const goods = [
...
+                <button
+                  data-cy="ClearButton"
+                  type="button"
+                  className="delete ml-3"
+                  onClick={this.removeGood}
+                />
+              </>
+            )
+            : 'No goods selected'}
+        </h1>
```
andrii-andriushchenko, line 53: you can move `h1` tag also into ternary operator, that makes your code more clear and change your fragment (`<>...</>`) , but up to you)

andrii-andriushchenko (changes requested):
Good job 🔥 
I left some comments for fix)

src/App.tsx
andrii-kurash: Thank's, I didn't see it👍

src/App.tsx
andrii-kurash: Good advice🔥

src/App.tsx
```diff
@@ -15,71 +16,87 @@ export const goods = [
...
-          </td>
-        </tr>
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
+                data-cy="Good"
+                key={good}
+                className={classNames(
+                  { 'has-background-success-light': good === selectedGood },
```
Anna-Tkachenko, line 63: you checked is current good selected several times
let's create separate function or boolean variable for this value

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #1267, opened 2023-01-30
solution

Kravchenko71213:
https://Kravchenko71213.github.io/react_goods-selector/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,95 @@ export const goods = [
...
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  buttonClick = (nameToGood: string) => {
```
ihor-soloviov, line 28: Not the best name for a button. The name should mean the action when the button is clicked.

src/App.tsx
```diff
@@ -15,71 +16,95 @@ export const goods = [
...
+    this.setState({
+      selectedGood: nameToGood,
```
ihor-soloviov, line 30: To my mind that's not the best variant haha. May 'nameOfGood' would be better?

src/App.tsx
```diff
@@ -15,71 +16,95 @@ export const goods = [
...
+                          -
+                        </button>
+                      )
+                      : (
+                        <button
+                          data-cy="AddButton"
+                          type="button"
+                          className="button"
+                          onClick={() => {
+                            this.buttonClick(good);
```
ihor-soloviov, line 91: This is a strange onClick making. There was the same with ( ' ' ) in parameters higher. I'd like to recommend you to make two functions for removing and adding goods.

ihor-soloviov (commented):
That is a good job! But to my mind there are some minor fixes you should do to make it excelent!

src/App.tsx
Anna-Tkachenko: agree, fix @ihor-soloviov comment

Anna-Tkachenko (changes requested):

src/App.tsx
```diff
@@ -15,71 +16,93 @@ export const goods = [
...
+                  <td>
+                    {isGoodSelected
+                      ? (
+                        <button
+                          data-cy="RemoveButton"
+                          type="button"
+                          className="button is-info"
+                          onClick={() => {
+                            this.clearSelected();
+                          }}
```
Anna-Tkachenko, line 78: check point 3 in checklist

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #1277, opened 2023-01-30
solution

imariash86:
[DEMO LINK](https://imariash86.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


imariash86:
> Fix DEMO LINK pls:)

thanks for info, fixed already

src/App.tsx
```diff
@@ -15,71 +16,118 @@ export const goods = [
...
-      </tbody>
-    </table>
-  </main>
-);
+const preparedGoods = goods.map((good, index) => (
+  {
+    name: good,
+    id: index,
+  }
+));
```
vanst1, line 24: checkList 4: NEVER EVER EVER use array index as a key prop

src/App.tsx
```diff
@@ -15,71 +16,118 @@ export const goods = [
...
+
+        <table className="table">
+          <tbody>
+            {preparedGoods.map(good => {
+              if (good.name === selectedGood) {
+                return (
+                  <tr
+                    data-cy="Good"
+                    className="has-background-success-light"
+                    key={good.id}
```
vanst1, line 84: don't use array index as a key prop

src/App.tsx
```diff
@@ -15,71 +16,118 @@ export const goods = [
...
+
+                    <td data-cy="GoodTitle" className="is-vcentered">
+                      {good.name}
+                    </td>
+                  </tr>
+                );
+              }
+
+              return (
+                <tr data-cy="Good" key={good.id}>
```
vanst1, line 107:
```suggestion
                <tr data-cy="Good" key={good.name}>
```
same here

src/App.tsx
```diff
@@ -15,71 +16,118 @@ export const goods = [
...
+                    >
+                      +
+                    </button>
+                  </td>
+
+                  <td data-cy="GoodTitle" className="is-vcentered">
+                    {good.name}
+                  </td>
+                </tr>
+              );
```
vanst1, line 125: try to avoid code duplication,
it would be best if you changed only button

src/App.tsx
```diff
@@ -15,71 +16,118 @@ export const goods = [
...
+                      >
+                        -
+                      </button>
+                    </td>
+
+                    <td data-cy="GoodTitle" className="is-vcentered">
+                      {good.name}
+                    </td>
+                  </tr>
+                );
```
vanst1, line 103: code duplication

vanst1 (changes requested):
Good try)
Some fix needed:)

vanst1 (approved):
Good job)

src/App.tsx
```diff
@@ -15,71 +16,107 @@ export const goods = [
...
+        {title}
+
+        <table className="table">
+          <tbody>
+            {goods.map(good => {
+              return (
+                <tr
+                  data-cy="Good"
+                  className={cd('', {
+                    'has-background-success-light': good === selectedGood,
```
Anna-Tkachenko, line 76: you checked is current good selected several times, create separate function for this purpose, don't repeat yourself

src/App.tsx
```diff
@@ -15,71 +16,107 @@ export const goods = [
...
+export class App extends Component<{}, States> {
+  state = {
+    selectedGood: 'Jam',
+  };
+
+  select = (good: string) => {
+    if (good === this.state.selectedGood) {
+      this.setState({
+        selectedGood: '',
+      });
```
Anna-Tkachenko, line 32: don't repeat yourself: let's check once is current good selected, according to those check will define the needed value that we need to set to the state, and call `setState` only once

imariash86: I didn't completely understand what you meant, but I did some changes, please check

Anna-Tkachenko: will be better to create some variable, like `newGood`, in which you'll store empty string or `good` value according to the check: is current good selected
and then only one time pass this variable to setState method, instead of copying your code and calling setState twice in one function

src/App.tsx
```diff
@@ -15,71 +16,112 @@ export const goods = [
...
+interface States {
+  selectedGood: string;
+}
+
+export class App extends Component<{}, States> {
+  state = {
+    selectedGood: 'Jam',
+  };
+
+  isCurrGoodSelected = (good: string) => {
```
Anna-Tkachenko, line 28: don't make function name shortened, will be better to use full naming

Anna-Tkachenko (approved):

PR #1294, opened 2023-02-02
gooodsSelector

Kukharuk:
[DEMO LINK](https://Kukharuk.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,81 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state = {
+    selectedGood: 'Jam',
+  };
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +15,81 @@ export const goods = [
...
-            </button>
-          </td>
+    return (
+      <main className="section container">
+        {selectedGood.length
+          ? (
+            <h1 className="title is-flex is-align-items-center">
+              {selectedGood}
+              {' '}
+              is selected
```
etojeDenys, line 37: we can write it in one line without `{' '}`

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -15,71 +15,81 @@ export const goods = [
...
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
VitaliyBondarenko1982, line 38: For what here eslint disable?
Looks like before button here was input without label.

src/App.tsx
VitaliyBondarenko1982: Please fix this comment

VitaliyBondarenko1982 (changes requested):

mykhalenych (approved):

PR #1306, opened 2023-02-08
react goods selector task completed

MykolaKaradzha:
[DEMO LINK](https://MykolaKaradzha.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import Classname from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +16,91 @@ export const goods = [
...
-            >
-              +
-            </button>
-          </td>
+        <table className="table">
+          <tbody>
+            {goods.map(
+              good => (
+                <tr
+                  data-cy="Good"
```
Viktor-Kost, line 61: add a key o an item when you render list of items [read](https://reactjs.org/docs/lists-and-keys.html#keys)

src/App.tsx
```diff
@@ -15,71 +16,91 @@ export const goods = [
...
-              -
-            </button>
-          </td>
+              className="delete ml-3"
+              onClick={clearSelectionHandler}
+            />
+          )}
+        </h1>
+      )
+      : <h1 className="title">No goods selected</h1>;
```
Viktor-Kost, line 50: either use this title part of code inside `return()` or create a separated file for this component

src/App.tsx
```diff
@@ -15,71 +16,91 @@ export const goods = [
...
+                          -
+                        </button>
+                      )
+                      : (
+                        <button
+                          data-cy="AddButton"
+                          type="button"
+                          className="button"
+                          onClick={() => {
+                            this.setState({ selectedGood: good });
```
Viktor-Kost, line 86: Although this click needs one line of code I would recommend to create a function for handling this click (similar to clearSelectionHandler). It makes your code more structured

Viktor-Kost (changes requested):

src/App.tsx
MykolaKaradzha: Yeap, my bad. Still forgetting about them.

Viktor-Kost: no, Math.random() is a bad idea the same as using index for your key. Use item text at least

MykolaKaradzha: Thanks, will read more about them in React docs.)

IvanFesenko (approved):

PR #1307, opened 2023-02-09
Finish task

U-Ra-All:
[DEMO LINK](https://U-Ra-All.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,71 +15,133 @@ export const goods = [
...
+    if (selectedGoodRow) {
+      selectedGoodRow.classList.add('has-background-success-light');
+      const button = selectedGoodRow.children[0]?.children[0];
+
+      if (button) {
+        (button as HTMLElement).dataset.cy = 'RemoveButton';
+        button.classList.add('is-info');
+        button.textContent = '-';
+      }
+    }
```
IvanFesenko, line 39: You shouldn't work with DOM!

IvanFesenko (changes requested):
You shouldn't work with DOM. 
Recomendation is use functional component.
If you have any questions - ask them in fe_chart

src/App.tsx
U-Ra-All: Ok, got it

src/App.tsx
```diff
@@ -15,71 +15,103 @@ export const goods = [
...
-          <td>
-            <button
-              data-cy="RemoveButton"
-              type="button"
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+            if ((selectedGood && selectedGood !== good) || !selectedGood) {
```
etojeDenys, line 61:
```suggestion
            if (!selectedGood || selectedGood !== good) {
```

src/App.tsx
```diff
@@ -15,71 +15,103 @@ export const goods = [
...
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+            if (selectedGood === good) {
```
etojeDenys, line 84: can we use here just `else`?

src/App.tsx
```diff
@@ -15,71 +15,103 @@ export const goods = [
...
-              data-cy="RemoveButton"
-              type="button"
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+            if ((selectedGood && selectedGood !== good) || !selectedGood) {
+              tableRow = (
+                <tr data-cy="Good" key={good}>
```
etojeDenys, line 63:
```suggestion
             return  (
                 <tr data-cy="Good" key={good}>
```

src/App.tsx
```diff
@@ -15,71 +15,103 @@ export const goods = [
...
-            </button>
-          </td>
+            if (selectedGood === good) {
+              tableRow = (
+                <tr
+                  className="has-background-success-light"
+                  data-cy="Good"
+                  key={good}
+                >
+                  <td>
```
etojeDenys, line 91:
```suggestion
              return (
                <tr
                  className="has-background-success-light"
                  data-cy="Good"
                  key={good}
                >
                  <td>
```

etojeDenys (changes requested):

src/App.tsx
U-Ra-All: We can use just return here
`return (
              <tr
                className="has-background-success-light"
                data-cy="Good"
                key={good}
              >`

etojeDenys (approved):
good job

U-Ra-All:
> good job

Thanks

PR #1310, opened 2023-02-11
solution

Deniskakaka:
https://deniskakaka.github.io/react_goods-selector/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +15,78 @@ export const goods = [
...
+type Props = {
+  select: string,
+  name: string,
+  switchSelect: (name: string) => void
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+const Element: React.FC<Props> = ({ select, name, switchSelect }) => {
+  const [check, setCheck] = useState<boolean>(select === name);
```
etojeDenys, line 25: do we actually need useState here, we cannot instead of using `check` variable just check if `select === name`?

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -15,71 +15,74 @@ export const goods = [
...
+      className={check ? 'has-background-success-light' : ''}
+    >
+      <td>
+        <button
+          data-cy={check ? 'RemoveButton' : 'AddButton'}
+          type="button"
+          className={check ? 'button is-info' : 'button'}
+          onClick={() => {
+            switchSelect(!check ? name : '');
+          }}
```
IvanFesenko, line 39: Create separate handler instead using anonymous function

src/App.tsx
```diff
@@ -15,71 +15,74 @@ export const goods = [
...
-            >
-              +
-            </button>
-          </td>
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+          <button
+            data-cy="ClearButton"
+            type="button"
+            className="delete ml-3"
+            onClick={() => setSelect('')}
```
IvanFesenko, line 68: Same here

IvanFesenko (approved):
Good job. Check comments for future

PR #1311, opened 2023-02-14
solution

TetianaKoval:
[DEMO LINK](https://TetianaKoval.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,88 @@ export const goods = [
...
-        </tr>
+    return (
+      <main className="section container">
+        {
+          selectedGood
+            ? (
+              <h1 className="title is-flex is-align-items-center">
+                {selectedGood}
+                {' '}
+                is selected
```
witflash, line 41: You can use string interpolation instead of this approach.

witflash (approved):
Good work 👍🏼

PR #1312, opened 2023-02-14
react_goods-selector

dima-korovai:
[DEMO LINK](https://dima-korovai.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,6 +2,8 @@ import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
+import classNames from 'classnames';
+
 export const goods = [
   'Dumplings',
   'Carrot',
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,71 +17,94 @@ export const goods = [
...
+type State = {
+  selectedGood: string;
+};
+
+export class App extends React.Component<{}, State> {
+  state = {
+    selectedGood: 'Jam',
+  };
+
+  HandlerSelectedGoods = () => {
```
polosanya, line 29:
```suggestion
  handleSelectedGood = () => {
```
Use camelCase. Also, name your function correctly
https://javascript.plainenglish.io/handy-naming-conventions-for-event-handler-functions-props-in-react-fc1cbb791364

src/App.tsx
```diff
@@ -15,71 +17,94 @@ export const goods = [
...
+    this.setState({ selectedGood: '' });
+  };
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+  HandlerRemoveGood = () => {
+    this.setState({ selectedGood: '' });
```
polosanya, line 34: Same here. And what's the difference between these two functions, if both of them do the same?
They are just resetting the selectedGood

src/App.tsx
```diff
@@ -15,71 +17,94 @@ export const goods = [
...
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+    return (
+      <main className="section container">
+        <h1 className="title is-flex is-align-items-center">
+          {!selectedGood
```
polosanya, line 43:
```suggestion
          {selectedGood
```
It's better to check if good is present in order not to confuse other developers

polosanya (changes requested):
Let's improve your code a bit)

src/App.tsx
dima-korovai: Thank you, Oleksandr.

OleziO (approved):

PR #1314, opened 2023-02-14
React_goods-selector

07Spirit:
[DEMO LINK](https://07Spirit.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +16,108 @@ export const goods = [
...
+  name: string;
+  selectedItem: boolean;
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    name: 'Jam',
+    selectedItem: true,
```
Anastasiia-Svintsova, line 27: You don't need 2 states here. Just use one that can be string or an empty for example
```suggestion
    selectedGood: 'Jam',
```

src/App.tsx
```diff
@@ -15,71 +16,108 @@ export const goods = [
...
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  controlEvent = (event: React.MouseEvent<HTMLButtonElement>) => {
+    const retouch = event.currentTarget.innerText === '+';
```
Anastasiia-Svintsova, line 31: You don't need to control the event here. Just change the state and that's it

src/App.tsx
```diff
@@ -15,71 +16,108 @@ export const goods = [
...
-              data-cy="RemoveButton"
-              type="button"
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+  clearAll = () => {
+    const button = document.querySelectorAll('.button');
+    const tr = document.querySelectorAll('tr');
```
Anastasiia-Svintsova, line 49: Everything is much simpler here.
You don't need all those difficult calculations like in js. Refactor all your functions, please 
```
clearState = () => {
};

File: src/App.tsx [hunk 4]
```diff
@@ -15,71 +16,108 @@ export const goods = [
...
-);
+  createMassiveElements = () => {
+    return (
+      goods.map((elem) => {
+        return (
+          <tr
+            data-cy="Good"
+            key={elem}
+            className={
+              classNames({ 'has-background-success-light': elem === 'Jam' })
```
Anastasiia-Svintsova, line 71: Why do you have this class only for Jam? 
```
elem === selectedGood;

File: src/App.tsx [hunk 5]
```diff
@@ -15,71 +16,108 @@ export const goods = [
...
-            Garlic
-          </td>
-        </tr>
-      </tbody>
-    </table>
-  </main>
-);
+  createMassiveElements = () => {
+    return (
+      goods.map((elem) => {
```
Anastasiia-Svintsova, line 65:
```suggestion
      goods.map((good) => {
```

src/App.tsx
```diff
@@ -15,71 +16,108 @@ export const goods = [
...
+
+  render() {
+    return (
+      <main className="section container">
+        {this.state.selectedItem
+          ? (
+            <h1 className="title is-flex is-align-items-center">
+              {`${this.state.name} is selected`}
+
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
Anastasiia-Svintsova, line 102: Don't disable eslint

Anastasiia-Svintsova (changes requested):
You seem to be a bit confused with React. If you have any questions ask them in the fe_chat, please

src/App.tsx
07Spirit: Without this "disable" the button requaires attribute lable. This line was write down before my text.

PR #1318, opened 2023-02-17
Develop

fenderukr:
https://fenderukr.github.io/react_goods-selector

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,9 @@
-import React from 'react';
+/* eslint-disable jsx-a11y/control-has-associated-label */
 import 'bulma/css/bulma.css';
+import React from 'react';
 import './App.scss';
 
-export const goods = [
+const goods = [
   'Dumplings',
   'Carrot',
   'Eggs',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,80 @@ export const goods = [
...
+  }
 
-        <tr data-cy="Good">
-          <td>
+  render() {
+    return (
+      <main className="section container">
+        {this.state.selectedGood ? (
+          <h1 className="title is-flex is-align-items-center">
+            {this.state.selectedGood}
```
agniya-i, line 45: it would be nice to decompose `this.state` to not it repeat each time

`const { selectedGood } = this.state;`

src/App.tsx
```diff
@@ -15,71 +16,80 @@ export const goods = [
...
+              >
+                <td>
+                  <button
+                    data-cy={this.state.selectedGood === good
+                      ? 'RemoveButton'
+                      : 'AddButton'}
+                    type="button"
+                    className={`button ${
+                      this.state.selectedGood === good && 'is-info'
+                    }`}
```
agniya-i, line 79: you can also use `classnames` if you need set class based on some condition

agniya-i (changes requested):

src/App.tsx
```diff
@@ -15,71 +17,82 @@ export const goods = [
...
+  render() {
+    const { selectedGood } = this.state;
 
-        <tr data-cy="Good">
-          <td>
+    return (
+      <main className="section container">
+        {selectedGood ? (
+          <h1 className="title is-flex is-align-items-center">
+            {this.state.selectedGood}
```
slavanosatskyi, line 48: you can use decomposed value here

src/App.tsx
```diff
@@ -15,71 +17,82 @@ export const goods = [
...
+export class App extends React.Component {
+  state: StateType = {
+    selectedGood: 'Jam',
+  };
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+  clearCartHandler = () => {
```
slavanosatskyi, line 29: handleClearCart

https://github.com/mate-academy/react_goods-selector/blob/master/checklist.md

slavanosatskyi (approved):
Good, though there are two comments, please check them

PR #1320, opened 2023-02-20
Goods selector second commit

Sergey-Mironenko:


Sergey-Mironenko:
https://Sergey-Mironenko.github.io/react_goods-selector/

src/App.tsx
```diff
@@ -15,71 +15,81 @@ export const goods = [
...
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string;
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends Component<{}, State> {
+  state = {
+    selectedGood: 'Jam',
```
OleziO, line 24: Better to set `goods[goods.indexOf('Jam')] || ""`. In your version, the app will work incorrectly when there is no Jam element in the array

src/App.tsx
```diff
@@ -15,71 +15,81 @@ export const goods = [
...
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  render() {
+    return (
+      <main className="section container">
+        <h1 className="title is-flex is-align-items-center">
+          {this.state.selectedGood !== ''
```
OleziO, line 31: Replace with `this.state.selectedGood`, it will be the same as `this.state.selectedGood !== ''`. Check other cases.

src/App.tsx
```diff
@@ -15,71 +15,81 @@ export const goods = [
...
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+        <table className="table">
+          <tbody>
+            {goods.map(good => {
+              return (
+                <tr
```
OleziO, line 51: Add key for mapped elements.

src/App.tsx
```diff
@@ -15,71 +15,81 @@ export const goods = [
...
+                  data-cy="Good"
+                  className={
+                    good === this.state.selectedGood
```
OleziO, line 54: Move this check to the variable in the map method.

OleziO (changes requested):
Good Job! Fix all comments for aprove

OleziO (approved):

PR #1322, opened 2023-02-21
solution

vladyslav-krainyk:
[DEMO LINK](https://vladyslav-krainyk.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,6 @@
-import React from 'react';
+import { Component } from 'react';
+import classNames from 'classnames';
+
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +17,93 @@ export const goods = [
...
-  </main>
-);
+                {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+                <button
+                  data-cy="ClearButton"
+                  type="button"
+                  className="delete ml-3"
+                  onClick={
+                    this.emptyState
+                  }
```
loralevitska, line 51: You don't need to add extra lines if you don't need to
```suggestion
                  onClick={this.emptyState}
```

src/App.tsx
```diff
@@ -15,71 +17,93 @@ export const goods = [
...
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Garlic
-          </td>
-        </tr>
-      </tbody>
-    </table>
-  </main>
-);
+                {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
loralevitska, line 44: Try to fix eslint errors and remove comments

src/App.tsx
```diff
@@ -15,71 +17,93 @@ export const goods = [
...
+                  className={classNames({
+                    'has-background-success-light': selectedGood === good,
+                  })}
+                >
+                  <td>
+                    <button
+                      data-cy={
+                        selectedGood === good
+                          ? 'RemoveButton'
+                          : 'AddButton'
```
loralevitska, line 76: move to variable before return in 64 line

src/App.tsx
```diff
@@ -15,71 +17,93 @@ export const goods = [
...
+                        'button',
+                        { 'is-info': selectedGood === good },
+                      )}
+                      onClick={() => {
+                        if (selectedGood === good) {
+                          this.emptyState();
+                        } else {
+                          this.setState({ selectedGood: good });
+                        }
+                      }}
```
loralevitska, line 89: Move to separate function  (transfer the function to a variable, and here only write the name of the function)

loralevitska (changes requested):
A little more and it will be completed

src/App.tsx
TarasHoliuk: not fixed
at least you can move `selectedGood === good` to a variable - you check it 4 times

TarasHoliuk (changes requested):
GJ :+1: 
only one thing remains to fix

src/App.tsx
```diff
@@ -15,71 +17,98 @@ export const goods = [
...
-              data-cy="RemoveButton"
-              type="button"
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+  checked = (selectedGood: string, good: string) => {
+    return selectedGood === good;
+  };
```
Viktor-Kost, line 43: why do you need a function for this? 
You can just create a variable after this line of code `const { selectedGood } = this.state;` at `render()` method
`const isChecked = selectedGood === good`

SanyaBratashchuk (approved):

PR #1329, opened 2023-02-25
did the task

vanyachyzh:
[DEMO LINK](https://vanyachyzh.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,71 +15,109 @@ export const goods = [
...
+
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+              <button
+                data-cy="ClearButton"
+                type="button"
+                className="delete ml-3"
+                onClick={() => {
+                  this.unsetCells();
+                  this.setState({ selectedGoods: '' });
+                }}
```
VitaliyBondarenko1982, line 90: Make function in component and pass reference here

src/App.tsx
```diff
@@ -15,71 +15,109 @@ export const goods = [
...
+    const cells = document.querySelectorAll('tr');
+    const buttons = document.querySelectorAll('button');
+
+    buttons.forEach((b: HTMLButtonElement) => {
+      // eslint-disable-next-line no-param-reassign
+      b.innerText = '+';
+    });
+
+    removeClass(buttons, 'is-info');
+    removeClass(cells, 'has-background-success-light');
```
VitaliyBondarenko1982, line 46: no need to work with DOM.
Work with state and decide which classes add to remove, which text set in button and so on conditionally in JSX.

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):

PR #1334, opened 2023-02-28
Solution

katerinalex:
[DEMO LINK](https://katerinalex.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -38,9 +38,7 @@ export class App extends React.Component<{}, State> {
           ? <h1 className="title">No goods selected</h1>
           : (
             <h1 className="title is-flex is-align-items-center">
-              {selectedGood}
-              {' '}
-              is selected
+              {`${selectedGood} is selected`}
 
               {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
               <button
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +15,89 @@ export const goods = [
...
-              -
-            </button>
-          </td>
+        {selectedGood === ''
+          ? <h1 className="title">No goods selected</h1>
+          : (
+            <h1 className="title is-flex is-align-items-center">
+              {selectedGood}
+              {' '}
+              is selected
```
polosanya, line 43:
```suggestion
              {`${selectedGood} is selected`}
```

src/App.tsx
```diff
@@ -15,71 +15,89 @@ export const goods = [
...
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+        <table className="table">
+          <tbody>
+            {goods.map((good: string) => {
+              const selected = good === selectedGood;
```
polosanya, line 58:
```suggestion
              const isSelected = good === selectedGood;
```
Name it like this, so it's clear that it is a boolean value

polosanya (changes requested):
Let's polish it up)

polosanya (approved):
Well done!

PR #1337, opened 2023-03-01
Solution

OksanaShvets1:
[DEMO LINK](https://OksanaShvets1.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +15,152 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+// export const App: React.FC = () => (
+//   <main className="section container">
```
tetiana-ivanchenko, line 19: dont left comments in your code

src/App.tsx
```diff
@@ -15,71 +15,152 @@ export const goods = [
...
+          {buttonClear}
+        </h1>
+
+        <table className="table">
+          <tbody>
+            {goods.map((good) => (
+              <tr
+                key={good}
+                data-cy="Good"
+                className={good === selectedGood
```
tetiana-ivanchenko, line 130: use classnames library

tetiana-ivanchenko (changes requested):

src/App.tsx
```diff
@@ -15,71 +16,84 @@ export const goods = [
...
+  };
+
+  handleClearSelection = () => {
+    this.setState({ selectedGood: '' });
+  };
+
+  render() {
+    const { selectedGood } = this.state;
+    const title = selectedGood ? `${selectedGood} is selected` : 'No goods selected';
+    const buttonClear = selectedGood && (
```
TarasHoliuk, line 39: You don't re-use it, so I don't see any reason to move this part of `JSX` to a variable
Just write it in place

src/App.tsx
```diff
@@ -15,71 +16,84 @@ export const goods = [
...
-            >
-              +
-            </button>
-          </td>
+    return (
+      <main className="section container">
+        <h1 className="title">
+          {title}
+          {buttonClear}
+        </h1>
```
TarasHoliuk, line 54: The clear button is not aligned:
image

TarasHoliuk (changes requested):
GJ :+1: 
But there is one thing that should be fixed

src/App.tsx
```diff
@@ -15,71 +16,82 @@ export const goods = [
...
-    <h1 className="title">No goods selected</h1>
+interface AppState {
+  selectedGood: string;
+}
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends Component<{}, AppState> {
+  state: AppState = {
+    selectedGood: 'Jam',
```
OleziO, line 25: Replace with `goods[goods.indexOf('Jam')] || goods[0]`

In your code - if you don't have the element 'Jam' on the goods array - it will make bugs.

OleziO (changes requested):

sTorba24 (approved):
Well done!

PR #1339, opened 2023-03-03
first try

RudBeee:
https://RudBeee.github.io/react_goods-selector/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,105 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string,
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state = {
+    selectedGood: 'Jam',
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  selected = (good1: string) => {
+    return this.state.selectedGood !== good1
+      ? this.setState({ selectedGood: good1 })
+      : this.setState({ selectedGood: '' });
+  };
 
-    <table className="table">
-      <tbody>
-        <tr data-cy="Good">
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+  deleteH1 = () => {
+    this.setState({ selectedGood: '' });
+  };
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+  render() {
+    const { selectedGood } = this.state;
 
-        <tr data-cy="Good" className="has-background-success-light">
-          <td>
-            <button
-              data-cy="RemoveButton"
-              type="button"
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+    return (
+      <main className="section container">
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Jam
-          </td>
-        </tr>
+        {selectedGood === ''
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +15,105 @@ export const goods = [
...
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string,
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state = {
+    selectedGood: 'Jam',
```
OleziO, line 24: Replace with `goods[goods.indexOf('Jam')] || goods[0]`.

About your code -> If you will not have the element 'Jam' - you will have some bugs.

src/App.tsx
```diff
@@ -15,71 +15,105 @@ export const goods = [
...
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+  deleteH1 = () => {
```
OleziO, line 33: Rename to `clearSelected`

src/App.tsx
```diff
@@ -15,71 +15,105 @@ export const goods = [
...
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Jam
-          </td>
-        </tr>
+        {selectedGood === ''
+          ? (<h1 className="title">No goods selected</h1>)
+          : (
+            <h1 className="title is-flex is-align-items-center">
+              {selectedGood}
```
OleziO, line 47: Use template strings. 

[More info.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

src/App.tsx
```diff
@@ -15,71 +15,105 @@ export const goods = [
...
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+              <button
```
OleziO, line 52: Use `` tag inside `<button>`, its better practice than setting the background image.

src/App.tsx
```diff
@@ -15,71 +15,105 @@ export const goods = [
...
+
+//     {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+//     <button
+//       data-cy="ClearButton"
+//       type="button"
+//       className="delete ml-3"
+//       onClick={this.deleteH1}
+//     />
+//   </h1>
+// )}
```
OleziO, line 119: Remove all comments from the code.

OleziO (changes requested):

src/App.tsx
```diff
@@ -15,71 +15,105 @@ export const goods = [
...
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  selected = (good1: string) => {
+    return this.state.selectedGood !== good1
```
OleziO, line 28: Just do `this.setState({ selectedGood: goodName });`

src/App.tsx
```diff
@@ -15,71 +15,105 @@ export const goods = [
...
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  selected = (good1: string) => {
```
OleziO, line 27: Rename to `selectGood`

OleziO (changes requested):

src/App.tsx
RudBeee: ok, but I think it's better to leave an empty line.
Like this goods[goods.indexOf('Jam')] || ''

src/App.tsx
RudBeee: but in the file it is not indicated that it is required to change the HTML

src/App.tsx
RudBeee: then how to change to empty string if element is already selected?

mykhalenych (approved):

PR #1340, opened 2023-03-03
solution

vlad-trundenko:
[DEMO LINK](https://vlad-trundenko.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -36,9 +36,7 @@ export class App extends Component<{}, State> {
       <main className="section container">
         {selectedGood.length ? (
           <h1 className="title is-flex is-align-items-center">
-            {selectedGood}
-            {' '}
-            is selected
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +16,87 @@ export const goods = [
...
-            Dumplings
-          </td>
-        </tr>
+    return (
+      <main className="section container">
+        {selectedGood.length ? (
+          <h1 className="title is-flex is-align-items-center">
+            {selectedGood}
+            {' '}
+            is selected
```
witflash, line 41: You can use string interpolation here:
```
```
It looks better.

witflash (changes requested):
Hey!
Your DEMO link doesn't work so we can't check your page behavior.
Could you check it? Maybe you didn't deploy it?

agniya-i (approved):

PR #1343, opened 2023-03-04
Solution

Vasyl-Pavlenko:
[DEMO LINK](https://vasyl-pavlenko.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,114 @@ export const goods = [
...
-            </button>
-          </td>
-
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Jam
-          </td>
-        </tr>
+              {`${selectedGood} is selected`}
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+              <button
```
OleziO, line 57: Add `` tag inside `<button>` (don't use a background image in this case), it's good practice, and its adds accessibility to your application.

OleziO (changes requested):

src/App.tsx
Vasyl-Pavlenko: I don't sure that I have understood you right.
1. Needs delete classes from button
2 Add tag img
3 Add some icon in the source of img
4 right my own class for button and img

That's right?

etojeDenys: nope, just rewrite button tag, instead of `<button/>` use `<button></button>`

etojeDenys (changes requested):
small improvement to fix the error with button

src/App.tsx
```diff
@@ -15,71 +16,115 @@ export const goods = [
...
+                data-cy="ClearButton"
+                type="button"
+                className="
+                  delete
+                  ml-3
+                "
+                onClick={this.removeGood}
+              >
+                <img src="" alt="" />
```
etojeDenys, line 65: it`s line redundant

etojeDenys (approved):
LGTM

PR #1347, opened 2023-03-06
initial commit

ihorivna007:
[DEMO LINK](https://ihorivna007.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +15,93 @@ export const goods = [
...
+  render() {
+    const { selectedGood } = this.state;
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+    return (
+      <main className="section container">
+        {(selectedGood !== '')
```
Volodymyr-Melnyk-Fron10der, line 36: `var !== ''` - it means value is not empty, which means value is not false, which means value is true
```suggestion
        {selectedGood
```

Volodymyr-Melnyk-Fron10der (approved):
Looks good in general👍 Just check my comment please to avoid the same mistakes in future

PR #1361, opened 2023-03-13
Solution

Dynamizx:
https://mate-academy.github.io/react_goods-selector/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -32,39 +32,37 @@ export class App extends React.Component<{}, State> {
 
     return (
       <main className="section container">
-        <h1 className={`title ${selectedGood ? 'is-flex is-align-items-center' : ''}`}>
-          <div>
-            {selectedGood ? (
-              <div>
-                {`${selectedGood} is selected`}
-                {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-                <button
-                  data-cy="ClearButton"
-                  type="button"
-                  className="delete ml-3"
-                  onClick={this.clearGoodsSelection}
-                />
-              </div>
-            ) : (
-              <div>No goods selected</div>
-            )}
-          </div>
-        </h1>
+        {selectedGood ? (
+          <h1 className={`title ${selectedGood ? 'is-flex is-align-items-center' : ''}`}>
+            {`${selectedGood} is selected`}
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+            <button
+              data-cy="ClearButton"
+              type="button"
+              className="delete ml-3"
+              onClick={this.clearGoodsSelection}
+            />
+          </h1>
+        ) : (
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +16,85 @@ export const goods = [
...
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+    return (
+      <main className="section container">
+        <h1 className={`title ${selectedGood ? 'is-flex is-align-items-center' : ''}`}>
+          {selectedGood ? (
+            <>
```
polosanya, line 37: You can use h1 instead of React fragment. In that case, you will not need to render different classes, you will just have two different h1

src/App.tsx
```diff
@@ -15,71 +16,85 @@ export const goods = [
...
+              {selectedGood}
+              {' '}
+              is selected
```
polosanya, line 40:
```suggestion
              {`${selectedGood} is selected`}
```

polosanya (approved):
Let's improve your code a bit)

polosanya (changes requested):
Comments above

src/App.tsx
```diff
@@ -15,71 +16,85 @@ export const goods = [
...
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+    return (
+      <main className="section container">
+        <h1 className={`title ${selectedGood ? 'is-flex is-align-items-center' : ''}`}>
+          <div>
+            {selectedGood ? (
+              <div>
```
polosanya, line 38:
```suggestion
        {selectedGood ? (
          <h1 className="title is-flex is-align-items-center">
```
Why did you add these divs? Just move the h1 inside

src/App.tsx
```diff
@@ -15,71 +16,85 @@ export const goods = [
...
+              </div>
+            ) : (
+              <div>No goods selected</div>
+            )}
+          </div>
+        </h1>
+        <table className="table">
+          <tbody>
+            {goods.map(good => {
+              const goodIsSelected = good === selectedGood;
```
polosanya, line 56:
```suggestion
              const isGoodSelected = good === selectedGood;
```
image

src/App.tsx
```diff
@@ -15,71 +16,83 @@ export const goods = [
...
-          </td>
+              className="delete ml-3"
+              onClick={this.clearGoodsSelection}
+            />
+          </h1>
+        ) : (
+          <h1 className={`title ${selectedGood ? 'is-flex is-align-items-center' : ''}`}>
+            No goods selected
+          </h1>
+        )}
```
anstsot, line 50: the previous comment means that you can have the check inside your `h1` tag:
```
<h1>
  {selectedGood ? (
    // ...your code
  ) : (
    // ...your code
  )}
</h1>
```

anstsot (approved):
Good work!

PR #1363, opened 2023-03-14
react_goods-selector was done

pushkarskiyrodion:
[DEMO LINK](https://pushkarskiyrodion.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -23,7 +23,7 @@ type State = {
 
 export class App extends React.Component<{}, State> {
   state: Readonly<State> = {
-    selectedGood: 'Jam',
+    selectedGood: goods[goods.indexOf('Jam')] || goods[0],
   };
 
   removeGood = () => {


src/App.tsx
```diff
@@ -15,71 +17,83 @@ export const goods = [
...
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
```
SanyaBratashchuk, line 57: add keys

SanyaBratashchuk (changes requested):

src/App.tsx
```diff
@@ -15,71 +17,95 @@ export const goods = [
...
+                data-cy="Good"
+                className={cn('', {
+                  'has-background-success-light': good === selectedGood,
+                })}
+                key={this.setKey()}
```
Volodymyr-Melnyk-Fron10der, line 73: sorry, didn't get what did you mean here))
```suggestion
                key={good}
```

Volodymyr-Melnyk-Fron10der (changes requested):
Hi!  Almost there!

1. why is this icon visible? there is nothing to close in that case
image

src/App.tsx
pushkarskiyrodion: I define a function that will set a unique ID for key prop. It call everytime when component rerender itself.
I did that, because I need to provide an id. And I decide do that. Is it okay ?

polosanya: Actually, you don't pass any value. Your function returns nothing. 
The key shouldn't be just a random value, it should be connected to an element of an array.
Please fix mentor's comment

polosanya (changes requested):
Here you can read about keys in React and best practices of it:
https://www.developerway.com/posts/react-key-attribute

If you have any questions, please ask them in fe_chat and tag mentors. It will be faster because your answers in GitHub will be visible to the mentor only after you re-request review 🙂

src/App.tsx
pushkarskiyrodion: Got it. Thanks a lot :)

src/App.tsx
```diff
@@ -15,71 +17,88 @@ export const goods = [
...
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    selectedGood: 'Jam',
```
OleziO, line 26: Replace with `goods[goods.indexOf('Jam') || goods[0]`

OleziO (changes requested):

src/App.tsx
```diff
@@ -15,71 +17,88 @@ export const goods = [
...
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+  removeGood = () => {
```
OlEzhA0, line 29:
```suggestion
  clearSelectedGood = () => {
```

src/App.tsx
```diff
@@ -15,71 +17,88 @@ export const goods = [
...
-            >
-              +
-            </button>
-          </td>
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
+                data-cy="Good"
+                className={cn('', {
```
OlEzhA0, line 63:
```suggestion
                className={cn({
```

OlEzhA0 (approved):
added few minor improvements
GJ

PR #1369, opened 2023-03-16
initial solution

Maks-Tsarenko:
[DEMO LINK](https://Maks-Tsarenko.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +15,88 @@ export const goods = [
...
-              className="button"
-            >
-              +
-            </button>
-          </td>
+              return (
+                <tr
+                  key={good}
+                  data-cy="Good"
+                  className={isSelected ? 'has-background-success-light' : ''}
```
loralevitska, line 61: use [classnames](https://www.npmjs.com/package/classnames) library to add classes conditionally

src/App.tsx
```diff
@@ -15,71 +15,88 @@ export const goods = [
...
+                        </button>
+                      )
+                      : (
+                        <button
+                          data-cy="AddButton"
+                          type="button"
+                          className="button"
+                          onClick={() => {
+                            this.setState({ selectedGood: good });
+                          }}
```
loralevitska, line 82: Move to separate function  (transfer the function to a variable, and here only write the name of the function). Before jsx after 57 line for example

loralevitska (approved):

PR #1373, opened 2023-03-17
goods selector implemented

vladyslav-mosor:
[DEMO LINK](https://vladyslav-mosor.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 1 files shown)


sergey-lipikhin (approved):

src/App.tsx
```diff
@@ -15,71 +16,91 @@ export const goods = [
...
-              -
-            </button>
-          </td>
+    return (
+      <main className="section container">
+        {!selectedGood
+          ? <h1 className="title">No goods selected</h1>
+          : (
+            <h1 className="title is-flex is-align-items-center">
+              {`${selectedGood} is selected`}
```
OleksiiBlezniuk, line 45: But it's the same html element, only classes and content is different

Let's use one `h1` and select text and classes conditionally via classnames

OleksiiBlezniuk (approved):

src/App.tsx
vladyslav-mosor: Yes, that`s good point. Fixed!

PR #1387, opened 2023-03-18
Goods selector

VadimDrobyazko:
[DEMO LINK](https://vadimdrobyazko.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
+// eslint-disable-next-line import/no-extraneous-dependencies
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +17,99 @@ export const goods = [
...
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string;
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    selectedGood: 'Jam',
```
OleziO, line 26: Replace with `goods[goods.indexOf('Jam')] || goods[0]`. This is a safer case because if there is such an element in the array, it will choose the first one.

src/App.tsx
```diff
@@ -15,71 +17,99 @@ export const goods = [
...
+        </h1>
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Jam
-          </td>
-        </tr>
+        <table className="table">
+          <tbody>
+            {goods.map(good => {
+              const selectedG = good === selectedGood;
```
OleziO, line 60: Rename variable, like `isSelected`

OleziO (changes requested):

mykhalenych (approved):

PR #1427, opened 2023-03-30
solution

gabriella135:
[DEMO LINK](https://gabriella135.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,93 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string;
+};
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +15,93 @@ export const goods = [
...
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
+                key={good}
+                data-cy="Good"
+                className={
+                  isGoodSelected(good)
+                    ? 'has-background-success-light'
+                    : ''
```
polosanya, line 69: It is better to use [classnames tool](https://www.npmjs.com/package/classnames) for that
https://medium.com/2359media/why-react-classnames-is-a-lifesaver-and-how-you-should-use-it-if-you-arent-already-b25a44ed5603

polosanya (approved):
Good job!

PR #1442, opened 2023-04-11
Solution

yana-kotsulym:
https://yana-kotsulym.github.io/react_goods-selector/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+// import { flowRight } from 'cypress/types/lodash';
+// import { log } from 'console';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+// import { flowRight } from 'cypress/types/lodash';
```
SerhiiKirik, line 4: Remove all comments
```suggestion
```

src/App.tsx
```diff
@@ -15,71 +17,101 @@ export const goods = [
...
+
+export class App extends React.Component<{}, State> {
+  state = {
+    selectedValue: 'Jam',
+  };
+
+  clearValue = () => {
+    this.setState({ selectedValue: '' });
+
+    document.querySelectorAll('tr').forEach(el => {
```
SerhiiKirik, line 32: why do you need DOM here?
Use React property

src/App.tsx
```diff
@@ -15,71 +17,101 @@ export const goods = [
...
+            </h1>
+          )
+          : <h1 className="title">No goods selected</h1>}
+
+        <table className="table">
+          <tbody>
+            {goods.map(item => (
+              <tr
+                data-cy="Good"
+              >
```
SerhiiKirik, line 92: Wrap if you have > 2 params || length > 80
```suggestion
              <tr data-cy="Good">
```

SerhiiKirik: Also, when you make a map - add key to the parent tag
```suggestion
              <tr key={item} data-cy="Good">
```

src/App.tsx
```diff
@@ -15,71 +17,101 @@ export const goods = [
...
+                    +
+                  </button>
+                </td>
+
+                <td data-cy="GoodTitle" className="is-vcentered">
+                  {item}
+                </td>
+              </tr>
+            ))}
+
```
SerhiiKirik, line 111:
```suggestion
```

src/App.tsx
SerhiiKirik: just use state for it & classname libs

SerhiiKirik (changes requested):
Well done 🔥 
Let's make some improvements 🧑‍💻

src/App.tsx
```diff
@@ -15,71 +16,88 @@ export const goods = [
...
+                      item === this.state.selectedValue
+                        ? 'RemoveButton'
+                        : 'AddButton'
+                    }
+                    type="button"
+                    className={classNames('button', {
+                      'is-info': item === this.state.selectedValue,
+                    })}
+                    onClick={(event) => {
+                      if (event.currentTarget.dataset.cy === 'AddButton') {
```
anstsot, line 80:
```suggestion
                      if (!this.state.selectedValue) {
```

anstsot (approved):
Good job, approved. I've left one comment for proper check in the button

PR #1443, opened 2023-04-12
Solution

emxm:
https://emxm.github.io/react_goods-selector/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,80 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+export class App extends React.Component {
+  state = {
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,71 +15,80 @@ export const goods = [
...
-      />
-    </h1>
+    return (
+      <main className="section container">
+        { selectedGood
+          ? (
+            <h1 className="title is-flex is-align-items-center">
+              { selectedGood }
+              {' '}
+              is selected
```
IvanFesenko, line 33:
```suggestion
              `${selectedGood } is selected`
```

IvanFesenko (changes requested):
Demo link isn't working

src/App.tsx
emxm: photo_2023-04-12_17-59-08

qoa11a (approved):

PR #1452, opened 2023-04-14
solution

toljan1:
[DEMO LINK](https://toljan1.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
@@ -29,10 +30,6 @@ export const goods = [
 export class App extends React.Component {
   state = {
     selectedGood: 'Jam',
-    selectedClass: 'has-background-success-light',
-    noOneSelected: 'No goods selected',
-    addButtonText: '+',
-    removeButtonText: '-',
   };
 
   handleAddFunction = (event: React.MouseEvent<HTMLElement>): void => {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +26,98 @@ export const goods = [
...
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+export class App extends React.Component {
+  state = {
+    selectedGood: 'Jam',
+    selectedClass: 'has-background-success-light',
+    noOneSelected: 'No goods selected',
+    addButtonText: '+',
+    removeButtonText: '-',
```
VitaliyBondarenko1982, line 35: redundant a lot of states.
enough to keep in state only selected good
classes, text content render conditionally

src/App.tsx
```diff
@@ -15,71 +26,98 @@ export const goods = [
...
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+      event.currentTarget.classList.add(this.state.selectedClass);
+      (target as HTMLButtonElement).innerText = this.state.removeButtonText;
```
VitaliyBondarenko1982, line 50: do not work with DOM properties.
render conditionally in JSX text content, classes and so on.
for managing classes you can use library [classnames](https://www.npmjs.com/package/classnames)

VitaliyBondarenko1982 (changes requested):
check tests
set right data-cy for remove button
Screenshot from 2023-04-14 12-52-16
Screenshot from 2023-04-14 12-50-58

etojeDenys (approved):
lgtm

PR #1457, opened 2023-04-16
Develop

zynkd:
[DEMO LINK](https://zynkd.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,84 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string;
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state: State = {
+    selectedGood: 'Jam',
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  handleClick(item: string) {
+    // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+    item === this.state.selectedGood
+      ? this.setState({
+        selectedGood: '',
+      })
+      : this.setState({
+        selectedGood: item,
+      });
+  }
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +15,84 @@ export const goods = [
...
-    </h1>
+  handleClick(item: string) {
+    // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+    item === this.state.selectedGood
+      ? this.setState({
+        selectedGood: '',
+      })
+      : this.setState({
+        selectedGood: item,
+      });
```
polosanya, line 35: You can move ternary operator inside setState in order not to disable eslint.
But actually, this check is redundant, as it's impossible to choose the same good if it's already selected. (You can only remove it in that case)

src/App.tsx
```diff
@@ -15,71 +15,84 @@ export const goods = [
...
+  }
 
-    <table className="table">
-      <tbody>
-        <tr data-cy="Good">
-          <td>
+  render() {
+    return (
+      <main className="section container">
+        {this.state.selectedGood ? (
```
polosanya, line 41: Better destructure selectedGood from state to make your code a bit cleaner
`const { selectedGood } = this.state;`

src/App.tsx
```diff
@@ -15,71 +15,84 @@ export const goods = [
...
               type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+              className="delete ml-3"
+              onClick={() => this.setState({
+                selectedGood: '',
+              })}
```
polosanya, line 51: Create a separate function to handle this click

src/App.tsx
```diff
@@ -15,71 +15,84 @@ export const goods = [
...
+                    }
+                    type="button"
+                    className={
+                      item === this.state.selectedGood
+                        ? 'button is-info'
+                        : 'button'
+                    }
+                    onClick={() => this.handleClick(item)}
+                  >
+                    {item === this.state.selectedGood ? '-' : '+'}
```
polosanya, line 85:
```suggestion
                    {item === this.state.selectedGood 
                      ? '-' 
                      : '+'
                    }
```

polosanya (changes requested):
Almost done!

zynkd:
Let me know if all checks out. Thanks!

ivanchuk-yurii (approved):

PR #1466, opened 2023-04-18
solution

TanyaS133:
[DEMO LINK](https://TanyaS133.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


VenomMB (approved):
Great Job

src/App.tsx
```diff
@@ -15,71 +15,73 @@ export const goods = [
...
-            </button>
-          </td>
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
+                data-cy="Good"
+                className={selectedGood === good
+                  ? 'has-background-success-light' : ''}
+                // key={Math.random()}
```
liliia-livandovskaya, line 57: Using Math.random() as a key is not recommended. Keys should be unique and stable.
So add a key attribute as `key={good}`

src/App.tsx
```diff
@@ -15,71 +15,73 @@ export const goods = [
...
+              >
+                <td>
+                  <button
+                    data-cy={selectedGood === good
+                      ? 'RemoveButton' : 'AddButton'}
+                    type="button"
+                    className={selectedGood === good
+                      ? 'button is-info' : 'button'}
+                    onClick={() => {
+                      this.setState({ selectedGood: good });
```
liliia-livandovskaya, line 67: Instead of using a single onClick handler, consider using separate handlers for adding and removing a good. This will make the code more readable and maintainable 📚

liliia-livandovskaya (changes requested):
Great job! Lets work on fix

liliia-livandovskaya (approved):

PR #1473, opened 2023-04-18
Solution

Zad0rnik:


KosolapovAnatolii (approved):
Perfect!

src/App.tsx
```diff
@@ -15,71 +16,81 @@ export const goods = [
...
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  handleClick = (good: string) => {
+    return this.state.selectedGood === good
```
liliia-livandovskaya, line 29: The `handleClick` method should only set the selected good instead of toggling it. Remove the conditional and simply update the state with the new good.

src/App.tsx
```diff
@@ -15,71 +16,81 @@ export const goods = [
...
-              type="button"
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+        <table className="table">
+          <tbody>
+            {goods.map(good => {
+              const isGoodSelected = good === selectedGood;
```
liliia-livandovskaya, line 57: add key attribute

liliia-livandovskaya (approved):

PR #1478, opened 2023-04-18
Develop

Vlad-Lischenovskiy:
[DEMO LINK](https://Vlad-Lischenovskiy.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -83,3 +83,4 @@ export class App extends React.Component<{}, State> {
     );
   }
 }
+// work
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,71 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+interface State {
+  selectedGood: string | null;
+}
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    selectedGood: 'Jam',
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  render() {
+    const { selectedGood } = this.state;
 
-    <table className="table">
-      <tbody>
-        <tr data-cy="Good">
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+    return (
+      <main className="section container">
+        {selectedGood
+          ? (
+            <h1 className="title is-flex is-align-items-center">
+              {`${selectedGood} is selected` }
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +15,72 @@ export const goods = [
...
+            <h1 className="title is-flex is-align-items-center">
+              {`${selectedGood} is selected` }
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+              <button
+                data-cy="ClearButton"
+                type="button"
+                className="delete ml-3"
+                onClick={() => {
+                  this.setState({ selectedGood: null });
+                }}
```
vksazonov, line 43: It's better to put it to the separate function and reuse in other cases
See Checklist 2

src/App.tsx
```diff
@@ -15,71 +15,72 @@ export const goods = [
...
+              >
+                <td>
+                  <button
+                    data-cy={selectedGood === good
+                      ? 'RemoveButton' : 'AddButton'}
+                    type="button"
+                    className={selectedGood === good
+                      ? 'button is-info' : 'button'}
+                    onClick={() => {
+                      this.setState({ selectedGood: good });
```
vksazonov, line 66: Same here

vksazonov (changes requested):

vksazonov (approved):
Good Job Mate!

src/App.tsx
```diff
@@ -15,71 +15,74 @@ export const goods = [
...
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  handleClear = () => this.setState({ selectedGood: '' });
```
liliia-livandovskaya, line 27: The `handleClear` function should set the `selectedGood` to `null` instead of an empty string (`''`).

or you need to change  type in the interface State to
  `selectedGood: string;`

src/App.tsx
```diff
@@ -15,71 +15,74 @@ export const goods = [
...
+                  <button
+                    data-cy={selectedGood === good
+                      ? 'RemoveButton'
+                      : 'AddButton'}
+                    type="button"
+                    className={selectedGood === good
+                      ? 'button is-info'
+                      : 'button'}
+                    onClick={() => {
+                      this.setState({ selectedGood: good });
```
liliia-livandovskaya, line 69: Instead of updating the state twice, you can use a ternary operator to update `selectedGood` based on its current value.

src/App.tsx
liliia-livandovskaya: agree, create a handler for this

liliia-livandovskaya (changes requested):
Great job! Pay attention to the comments

src/App.tsx
liliia-livandovskaya: not fixed, create a handler for `this.setState({ selectedGood: good })`

liliia-livandovskaya (changes requested):
Almost done, pay attention to the comments

liliia-livandovskaya (approved):

PR #1492, opened 2023-04-19
goods solution

AP-Solution:
[DEMO LINK](https://ap-solution.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,74 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+interface State {
+  selectedGood: string;
+}
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state = {
+    selectedGood: 'Jam',
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +15,74 @@ export const goods = [
...
-    </table>
-  </main>
-);
+              return (
+                <tr
+                  data-cy="Good"
+                  className={isGoodSelected
+                    ? 'has-background-success-light'
+                    : ''}
+                  key={good}
```
roshirin, line 59: This will generate non-unique keys if we have several goods with the same name.

src/App.tsx
```diff
@@ -15,71 +15,74 @@ export const goods = [
...
+                      type="button"
+                      className={isGoodSelected
+                        ? 'button is-info'
+                        : 'button'}
+                      onClick={() => (isGoodSelected
+                        ? this.handleClear()
+                        : this.setState({ selectedGood: good }))}
+                    >
+                      {isGoodSelected ? '-' : '+'}
+                    </button>
```
roshirin, line 75: I think it would be nice to handle this logic in a separate function

roshirin (approved):
Nice solution!

I have two suggestions about it, check the comments plz.

src/App.tsx
```diff
@@ -15,71 +15,74 @@ export const goods = [
...
+                      data-cy={isGoodSelected
+                        ? 'RemoveButton'
+                        : 'AddButton'}
+                      type="button"
+                      className={isGoodSelected
+                        ? 'button is-info'
+                        : 'button'}
+                      onClick={() => (isGoodSelected
+                        ? this.handleClear()
+                        : this.setState({ selectedGood: good }))}
```
liliia-livandovskaya, line 72: create a handler for this

liliia-livandovskaya (approved):

PR #1493, opened 2023-04-19
goods selector

glots-a:
[DEMO LINK](https://glots-a.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,80 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  isChecked: boolean;
+  selectedGood: string;
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component <{}, State> {
+  state = {
+    isChecked: true,
+    selectedGood: 'Jam',
+  };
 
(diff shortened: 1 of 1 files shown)


MylaCrystal (changes requested):
Check your code: it turns out that you have only one button - you can choose, but cannot remove
your class (instead, only a + to - change).
And you have 2 different data-cy for +/- button

src/App.tsx
```diff
@@ -15,71 +16,87 @@ export const goods = [
...
-  </main>
-);
+            {goods.map((item: string) => {
+              const isItemSelected = item === selectedGood;
+
+              return (
+                <tr
+                  key={item}
+                  data-cy="Good"
+                  className={isItemSelected
```
SerhiiKirik, line 74: use `classnames` lib for it

src/App.tsx
```diff
@@ -15,71 +16,87 @@ export const goods = [
...
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+  clearButton = () => (
```
SerhiiKirik, line 37:
```suggestion
  handleClear = () => (
```

SerhiiKirik (changes requested):
Great 🔥 
Let's fix some bugs 🧑‍💻 
i can't uncheck the goods
image

PR #1494, opened 2023-04-19
goods selector solution

VladislavaT:
[DEMO LINK](https://VladislavaT.github.io/react_goods-selector/)


Hamuud (approved):
I hope that one day you will create an app on ⚛️React to record all the goods and it will be called: "Yummy Dumplings"😋

jstmpelowycz (approved):

PR #1506, opened 2023-04-25
solution

AnastassiaH:
[DEMO LINK](https://anastassiah.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -76,6 +76,9 @@ export class App extends Component<{}, State> {
                         data-cy="RemoveButton"
                         type="button"
                         className="button is-info"
+                        onClick={() => {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +19,77 @@ export const goods = [
...
+                        +
+                      </button>
+                    )
+                    : (
+                      <button
+                        data-cy="RemoveButton"
+                        type="button"
+                        className="button is-info"
+                      >
+                        -
```
vlad-demchuk, line 80: `-` should remove selected item from state

vlad-demchuk (changes requested):

vlad-demchuk (approved):

PR #1507, opened 2023-04-26
some_goods_available_for_selection

oleksiy-uchaev:
[DEMO LINK](https://oleksiy-uchaev.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,6 @@
+/* eslint-disable react/jsx-one-expression-per-line */
 import React from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,4 +1,6 @@
+/* eslint-disable react/jsx-one-expression-per-line */
```
polosanya, line 1: Don't disable the eslint without extra need and make sure that the code follows this rule. It will improve the code readability

src/App.tsx
```diff
@@ -15,71 +17,109 @@ export const goods = [
...
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state = {
+    selectedGoods: 'Jam',
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+  get oneGood(): string {
```
polosanya, line 29:
```suggestion
  getSelectedGood(): string {
```
Actually, there is no need to use this method, you can just take selectedGood from state

src/App.tsx
```diff
@@ -15,71 +17,109 @@ export const goods = [
...
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGoods: string;
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state = {
+    selectedGoods: 'Jam',
```
polosanya, line 26:
```suggestion
    selectedGood: 'Jam',
```
Only one good can be selected at a time

polosanya (changes requested):
Good job! Let's improve your code a bit)

agniya-i (approved):

PR #1509, opened 2023-04-27
Solution

Zibi95:
[DEMO LINK](https://Zibi95.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,6 +2,10 @@ import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
+type AppState = {
+  selectedGood: string;
+};
+
 export const goods = [
   'Dumplings',
   'Carrot',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +19,93 @@ export const goods = [
...
+        <table className="table">
+          <tbody>
+            {goods.map((good) => {
+              const isSelected = selectedGood === good;
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Jam
-          </td>
-        </tr>
+              const button = isSelected
```
Radoslaw-Czerniawski, line 62: Same here, you could have a separate component that would handle displaying either + or - button, it would receive `isSelected={selectedGood === good}` and  `handleClick={// either clearGood or () => setState}` and then you could just map the values without additional logic like declaring variables

Radoslaw-Czerniawski (changes requested):
It works but I would like you to do some refactoring. One question, are you required to use class components? If not then prefer to use functional components, class components are old and less preferable way of working with react (they are still supported for legacy reasons).

src/App.tsx
Zibi95: image

Zibi95:
There is state and for now it wasn't represented how to use it with functional components.

src/App.tsx
Radoslaw-Czerniawski: Aight didn't see that part about not creating additional components.

Radoslaw-Czerniawski (approved):
Well done :D

PR #1515, opened 2023-05-01
solution

Oleh-Rybachuk:
https://oleh-rybachuk.github.io/react_goods-selector/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,93 @@ export const goods = [
...
-          </td>
+    this.setState({
+      selectedGood: word,
+    });
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+    if (data === 'RemoveButton') {
```
YegorKorenkov, line 35: now, when clear a good, it is still called setState(row 31-33). Try to use if..else construction or use handleClearButton function for remove button if good selected

YegorKorenkov (changes requested):
Please fix minor mistake

nazarmatsevych (approved):

PR #1535, opened 2023-05-10
initial version

Ozwiena:
[DEMO LINK](https://ozwiena.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable react/prefer-stateless-function */
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +16,102 @@ export const goods = [
...
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+export class App extends React.Component<Props, State> {
+  state: Readonly<State> = {
+    selectedGood: 'Jam',
+    isSelected: true,
```
d-wasilewski, line 29: `isSelected` is redundant. You can just check if selectedGood is null, you don't need additional variable to track this state

d-wasilewski (changes requested):

src/App.tsx
```diff
@@ -15,71 +16,100 @@ export const goods = [
...
-        </tr>
+    return (
+      <main className="section container">
+        <h1 className={
+          selectedGood === null
+            ? 'title'
+            : 'title is-flex is-align-items-center'
+        }
+        >
+          {selectedGood === null
```
d-wasilewski, line 49: You don't need to explicitly compare it to null, undefined, empty string etc
```suggestion
          {!selectedGood
```

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #1538, opened 2023-05-11
Solution

NadieinOleh:
https://NadieinOleh.github.io/react_goods-selector/


witflash (changes requested):
Good work!

But the last thing.
The `minus` button should deselect the current item:
image

qoa11a (approved):

PR #1540, opened 2023-05-12
Solution

SuaBy514:
[DEMO LINK](https://suaby514.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,6 +2,10 @@ import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
+interface State {
+  selectedGood: string | null,
+}
+
 export const goods = [
   'Dumplings',
   'Carrot',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -2,6 +2,10 @@ import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
+interface State {
+  selectedGood: string | null,
```
kasiakubisiak, line 6: string that is blank will return false when converted to boolean, so you don't need to check for null here
```suggestion
  selectedGood: string,
```

src/App.tsx
```diff
@@ -15,71 +19,76 @@ export const goods = [
...
+    selectedGood: 'Jam',
+  };
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+  handleButton(relatedGood: string | null) {
+    return this.setState({ selectedGood: relatedGood });
+  }
```
kasiakubisiak, line 29: maybe consider adding a new button that will delete goods?
`this.setState({ selectedGood: '' });`

src/App.tsx
```diff
@@ -15,71 +19,76 @@ export const goods = [
...
+  render() {
+    const { selectedGood } = this.state;
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Jam
-          </td>
-        </tr>
+    return (
+      <main className="section container">
+        {!selectedGood
```
kasiakubisiak, line 36: It isn't really mistake here but suggestion - maybe let's reverse it, so firstly check for `selectedGood` not `!selectedGood`, its more intuitive

src/App.tsx
SuaBy514: I don't see the point of doing this, it is redundant as the existing function handles that perfectly fine

kasiakubisiak: I just thought of making it more descriptive to read code, but you may as well change `handleButton` name to something more descriptive, e.g. `toggleSelectedGood`

kasiakubisiak (changes requested):

kasiakubisiak (approved):

PR #1562, opened 2023-05-29
Done

petrolozynskyi:
[DEMO LINK](https://petrolozynskyi.github.io/react_goods-selector/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,3 +1,7 @@
 iframe {
   display: none;
 }
+
+.has-background-success-light {
+  background-color: #3e8ed0 !important;
+}
--- a/src/App.scss
+++ b/src/App.scss
@@ -4,4 +4,7 @@ iframe {
 
 .has-background-success-light {
   background-color: #3e8ed0 !important;
+  outline: none;
+  border: none;
 }
+
--- a/src/App.scss
+++ b/src/App.scss
@@ -2,9 +2,3 @@ iframe {
   display: none;
 }
 
-.has-background-success-light {
-  background-color: #3e8ed0 !important;
-  outline: none;
-  border: none;
-}
-
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -28,8 +28,7 @@ export class App extends React.Component<{}, State> {
     const { selectedGood } = this.state;
 
     this.setState({
-      selectedGood: selectedGood !== clickedGood
-        ? clickedGood : undefined,
+      selectedGood: selectedGood !== clickedGood ? clickedGood : undefined,
     });
   };
 
@@ -41,7 +40,6 @@ export class App extends React.Component<{}, State> {
         {selectedGood ? (
           <h1 className="title is-flex is-align-items-center">
             {`${selectedGood} is selected`}
-
             <button
               data-cy="ClearButton"
               type="button"
@@ -57,18 +55,16 @@ export class App extends React.Component<{}, State> {
 
         <table className="table">
           <tbody>
-            {goods.map(good => (
-              <tr
-                key={good}
-                data-cy="Good"
-                className={selectedGood === good
-                  ? 'has-background-success-light' : ''}
-              >
+            {goods.map((good) => (
+              <tr key={good} data-cy="Good">
                 <td>
                   <button
                     data-cy="AddButton"
                     type="button"
-                    className="button"
+                    className={`button ${
+                      selectedGood === good
+                        ? 'has-background-success-light' : ''
+                    }`}
                     onClick={() => this.handleButtonClick(good)}
                   >
                     {selectedGood === good ? '-' : '+'}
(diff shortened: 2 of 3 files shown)


src/App.tsx
```diff
@@ -15,71 +15,93 @@ export const goods = [
...
+                className={selectedGood === good
+                  ? 'has-background-success-light' : ''}
+              >
+                <td>
+                  <button
+                    data-good={good}
+                    data-cy="AddButton"
+                    type="button"
+                    className="button"
+                    onClick={this.handleButtonClick}
```
ivbutsykin, line 75: Default approach in React, it's pass `good` here, so:
```suggestion
                    onClick={() => this.handleButtonClick(good)}
```
In such way, you won't need `data-good`

src/App.tsx
```diff
@@ -15,71 +15,93 @@ export const goods = [
...
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  handleButtonClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+    const { selectedGood } = this.state;
+    const clickedGood = event.currentTarget.dataset.good;
```
ivbutsykin, line 29: Rewrite it, to receive `good` in `handleButtonClick`

src/App.tsx
```diff
@@ -15,71 +15,93 @@ export const goods = [
...
-              className="button"
-            >
-              +
-            </button>
-          </td>
+    if (selectedGood === clickedGood) {
+      this.setState({ selectedGood: undefined });
+    } else {
+      this.setState({ selectedGood: clickedGood });
+    }
```
ivbutsykin, line 35:
```suggestion
   this.setState({ selectedGood: selectedGood !== clickedGood ? clickedGood : undefined });
```

ivbutsykin (changes requested):

src/App.tsx
```diff
@@ -15,71 +15,90 @@ export const goods = [
...
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string | undefined;
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
+    selectedGood: 'Jam',
```
OleziO, line 24: Replace with `goods[goods.indexOf('Jam')] || goods[0]`

It's a more safe solution.

src/App.tsx
```diff
@@ -15,71 +15,90 @@ export const goods = [
...
-            Jam
-          </td>
-        </tr>
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
+                key={good}
+                data-cy="Good"
+                className={selectedGood === good
```
OleziO, line 64: Create a variable from this to avoid code duplication.

src/App.tsx
```diff
@@ -15,71 +15,90 @@ export const goods = [
...
-            Garlic
-          </td>
-        </tr>
-      </tbody>
-    </table>
-  </main>
-);
+              <td data-cy="GoodTitle" className="is-vcentered">
+                Garlic
+              </td>
```
OleziO, line 97: Remove this

OleziO (changes requested):
Fix button color when good is chosen
image

Delete the last item, it is a duplication of the previous and it is n clickable.

OleziO (changes requested):
Demo:
image
Design:
image

src/App.tsx
```diff
@@ -15,71 +15,76 @@ export const goods = [
...
-          </td>
-        </tr>
-      </tbody>
-    </table>
-  </main>
-);
+                  <td data-cy="GoodTitle" className="is-vcentered">
+                    {good}
+                  </td>
+                </tr>
```
ivbutsykin, line 82: Please, double check your styles. At that moment I still see white background on active item:
image

I suggest:
1. Add `has-background-success-light` class for `<tr>`
2. Your button will have `is-info` class depends on `isSelected`
3. You can remove `minusStyle` because you can get same result without it

src/App.scss
```diff
@@ -1,3 +1,10 @@
 iframe {
   display: none;
 }
+
+.has-background-success-light {
+  background-color: #3e8ed0 !important;
+  outline: none;
+  border: none;
+}
```
ivbutsykin, line 9: Hm, no need to change this class. For blue button you have `is-info` class, for green background - `.has-background-success-light`.

ivbutsykin (changes requested):
Please, fix active item background. If you have any questions - ask them in fe_chat

src/App.tsx
```diff
@@ -57,22 +57,20 @@ export class App extends React.Component<{}, State> {
           <tbody>
             {goods.map((good) => {
               const isSelected = selectedGood === good;
-              const buttonClassName = `button ${
-                isSelected ? 'has-background-success-light' : ''
-              }`;
-              const minusStyle = isSelected ? { color: 'white' } : {};
+              const trClassName = isSelected
+                ? 'has-background-success-light' : '';
+              const buttonClassName = `button ${isSelected ? 'is-info' : ''}`;
```
TarasHoliuk, line 62: just FYI - it's better to use libraries as `classnames` for conditional class names

TarasHoliuk (approved):

PR #1566, opened 2023-05-31
Develop

generatedman:
[DEMO LINK](https://generatedman.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
-import { type } from 'os';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +15,68 @@ export const goods = [
...
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+              <button
+                data-cy="ClearButton"
+                type="button"
+                className="delete ml-3"
+                onClick={() => this.setState({ selectedGood: '' })}
```
dinaminahi, line 45: It is better to create  a separate function because using an arrow function in render creates a new function each time the component renders, which may break optimizations based on strict identity comparison

But it is ok to use it, just when you can avoid it, just avoid

src/App.tsx
```diff
@@ -15,71 +15,68 @@ export const goods = [
...
+                  : ''}
+              >
+                <td>
+                  <button
+                    data-cy="AddButton"
+                    type="button"
+                    className={selectedGood === good
+                      ? 'button is-info'
+                      : 'button'}
+                    onClick={() => this.setState({ selectedGood: good })}
```
dinaminahi, line 66: So in the task, it is said that clear button, should clear the selection, but yours instead is setting selection, + and - are doing the same

Think how can you fix this

dinaminahi (changes requested):
Good job! But there is one thing that we need to fix

src/App.tsx
```diff
@@ -15,71 +16,83 @@ export const goods = [
...
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  handleClickOnButton = (good = '') => (
```
naziamoff, line 28: this comment is more grammar-related, but still)
Not a mistake in your case, but more ergonomic this way
```suggestion
  handleButtonClick = (good = '') => (
```

naziamoff (approved):
awesome job 🔥

PR #1567, opened 2023-05-31
react class App  was added

Olena-Shevchenko:
[DEMO LINK](https://Olena-Shevchenko.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,85 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string,
+};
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +15,85 @@ export const goods = [
...
-          </td>
-        </tr>
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+              <button
+                data-cy="ClearButton"
+                type="button"
+                className="delete ml-3"
+                onClick={() => {
+                  this.setState({ selectedGood: '' });
+                }}
```
polosanya, line 44: It would be great to create a separate function to handle this click in order not to duplicate yourself
image

polosanya (changes requested):
Almost done!

PR #1571, opened 2023-06-06
done

victor-buldenko:
[DEMO LINK](https://victor-buldenko.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,81 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+export class App extends React.Component {
+  state = {
+    cellCheck: false,
+    selectedGood: 'Jam',
+  };
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+  render() {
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +15,81 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+export class App extends React.Component {
+  state = {
+    cellCheck: false,
```
SanyaBratashchuk, line 20: There is no need in this property, you can just check if selectedGood contains something

src/App.tsx
```diff
@@ -15,71 +15,81 @@ export const goods = [
...
+        )}
+        <table className="table">
+          <tbody>
+            {goods.map((item) => {
+              return (
+                <tr
+                  data-cy="Good"
+                  key={item}
+                  className={
+                    selectedGood === item ? 'has-background-success-light' : ''
```
SanyaBratashchuk, line 54: It is better to use classnames or clsx library for conditional classNames

SanyaBratashchuk (changes requested):

TarasHoliuk (approved):

PR #1572, opened 2023-06-07
1st version

Grechana:
https://Grechana.github.io/react_goods-selector/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +15,79 @@ export const goods = [
...
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string,
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends Component <{}, State> {
+  state = {
+    selectedGood: 'Jam',
```
OleziO, line 24: Replace with `goods[goods.indexOf('Jam')] || goods[0]`. It's more safe variant

OleziO (changes requested):

src/App.tsx
```diff
@@ -15,71 +15,77 @@ export const goods = [
...
+              >
+                <td>
+                  <button
+                    type="button"
+                    onClick={() => {
+                      this.setState((prevState) => ({
+                        selectedGood:
+                        prevState.selectedGood === good ? '' : good,
+                      }));
+                    }}
```
polosanya, line 70: It would be great to create a separate function to handle this click

src/App.tsx
```diff
@@ -15,71 +15,77 @@ export const goods = [
...
-      </tbody>
-    </table>
-  </main>
-);
+        <table className="table">
+          <tbody>
+            { goods.map(good => (
+              <tr
+                data-cy="Good"
+                className={(selectedGood === good)
```
polosanya, line 58: Don't forget about key attribute
https://www.developerway.com/posts/react-key-attribute

src/App.tsx
```diff
@@ -15,71 +15,77 @@ export const goods = [
...
+                      }));
+                    }}
+                    className={(selectedGood === good)
+                      ? 'button is-info'
+                      : 'button'}
+                    data-cy={(selectedGood === good)
+                      ? 'RemoveButton'
+                      : 'AddButton'}
+                  >
+                    { (selectedGood === good) ? '-' : '+'}
```
polosanya, line 78: Nice. But actually you can just render two different buttons. It's up to you

polosanya (changes requested):
Almost perfect!

src/App.tsx
```diff
@@ -15,71 +15,79 @@ export const goods = [
...
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  handler = (good:string) => {
```
lerastarynets, line 27: Consider renaming the 'handler' function to a more descriptive name, such as 'handleSelectGood'.

lerastarynets (changes requested):
You are almost there) Only left one important comment from my side and also remember fixing all comments from the previous review

src/App.tsx
OleziO: Not resolved

OleziO (changes requested):

Volodymyr-Melnyk-Fron10der (approved):

PR #1575, opened 2023-06-09
done

VladyslavKucher:
https://VladyslavKucher.github.io/react_goods-selector/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +15,84 @@ export const goods = [
...
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  handler = (good = '') => (
```
polosanya, line 27: Consider renaming 'handler' to something more descriptive, like 'handleSelection'.

src/App.tsx
```diff
@@ -15,71 +15,84 @@ export const goods = [
...
+                <td>
+                  <button
+                    data-cy={
+                      this.state.selectedGood === good
+                        ? 'RemoveButton'
+                        : 'AddButton'
+                    }
+                    type="button"
+                    className={`button ${this.state.selectedGood === good ? 'is-info' : ''}`}
+                    onClick={() => {
```
polosanya, line 76: Remove the inline function and replace it with the previously suggested 'handleSelection' method, passing 'good' as a parameter.

src/App.tsx
```diff
@@ -15,71 +15,84 @@ export const goods = [
...
+                  this.state.selectedGood
+                  === good
+                    ? 'has-background-success-light'
+                    : ''
+                }
+              >
+                <td>
+                  <button
+                    data-cy={
+                      this.state.selectedGood === good
```
polosanya, line 70: Use destructured value everywhere

polosanya (changes requested):
Almost perfect!

src/App.tsx
```diff
@@ -15,71 +15,85 @@ export const goods = [
...
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
```
Anastasiia-Svintsova, line 46: Try not to disable eslint

src/App.tsx
```diff
@@ -15,71 +15,85 @@ export const goods = [
...
+              <button
+                data-cy="ClearButton"
+                type="button"
+                className="delete ml-3"
+                onClick={this.handleSelection()}
```
Anastasiia-Svintsova, line 51: When you assign an event handler to a component's onClick prop, you need to pass a reference to the handler function itself, rather than invoking it immediately.
```suggestion
                onClick={this.handleSelection}
```

src/App.tsx
```diff
@@ -15,71 +15,85 @@ export const goods = [
...
+                  <button
+                    data-cy={
+                      selectedGood === good
+                        ? 'RemoveButton'
+                        : 'AddButton'
+                    }
+                    type="button"
+                    className={`button ${selectedGood === good ? 'is-info' : ''}`}
+                    onClick={
+                      selectedGood === good
```
Anastasiia-Svintsova, line 80: You check if `selectedGood === good` not once. Move it to a variable, please

src/App.tsx
```diff
@@ -15,71 +15,85 @@ export const goods = [
...
+                        ? this.removeSelection()
+                        : this.handleSelection(good)
+                    }
+                  >
+                    {selectedGood === good ? '-' : '+'}
```
Anastasiia-Svintsova, line 85: In my opinion, it is preferable to render only two distinct buttons rather than performing multiple checks repeatedly.
```
{selectedGood === good ? (
  <button ...
) : ( 
  <button ...
)}

Anastasiia-Svintsova (changes requested):
Almost done! Let's add some fixes

Volodymyr-Melnyk-Fron10der (approved):

PR #1576, opened 2023-06-09
add task solution

NikaNika12:
https://NikaNika12.github.io/react_goods-selector/


demetrxx (approved):
Generally, nice work! 

You could have used only 1 state though.
Instead of tracking isSelected state, you could have set selectedGood state to null if no good is selected. And then conditionally render title on the basis of this state: selectedGood ? a : b.

PR #1579, opened 2023-06-12
Develop

serentinos:
[DEMO LINK](https://serentinos.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
 import { Component, ReactNode } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +16,87 @@ export const goods = [
...
+            {goodsList.map(item => {
+              const isSelected = item === selectedGood;
+
+              return (
+                <tr
+                  key={item}
+                  data-cy="Good"
+                  className={isSelected
+                    ? 'has-background-success-light'
+                    : undefined}
```
naziamoff, line 78: [classnames library](https://www.npmjs.com/package/classnames)

src/App.tsx
```diff
@@ -15,71 +16,87 @@ export const goods = [
...
+                >
+                  <td>
+                    <button
+                      data-cy={isSelected ? 'RemoveButton' : 'AddButton'}
+                      type="button"
+                      onClick={() => this.handleGoodsButton(item)}
+                      className={`button ${isSelected && 'is-info'}`}
```
naziamoff, line 85: https://www.npmjs.com/package/classnames

naziamoff (changes requested):
good job 👍 
But the time has come to get acquainted with the classnames library. You will like it

src/App.tsx
```diff
@@ -1,6 +1,8 @@
-import React from 'react';
+/* eslint-disable jsx-a11y/control-has-associated-label */
```
naziamoff, line 1: it is overall a bad practice to disable linting rules. But if you really need to do it - disable it only for the line where you need it instead of commenting the whole file
```suggestion
```

src/App.tsx
```diff
@@ -15,71 +17,92 @@ export const goods = [
...
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  handleResetSelected = () => {
+    this.setState({ selectedGood: '' });
+  };
```
naziamoff, line 31:
```suggestion
  clearSelection = () => {
    this.setState({ selectedGood: '' });
  };
```

src/App.tsx
```diff
@@ -15,71 +17,92 @@ export const goods = [
...
+          <tbody>
+            {goodsList.map(item => {
+              const isSelected = item === selectedGood;
+              const trSelectedLight = classNames({
+                'has-background-success-light': isSelected,
+              });
+              const buttonSelectedBackground = classNames({
+                button: true,
+                'is-info': isSelected,
+              });
```
naziamoff, line 78: you should do it where you assign attributes, no need to create a variable for that

src/App.tsx
```diff
@@ -15,71 +17,92 @@ export const goods = [
...
+
+              return (
+                <tr
+                  key={item}
+                  data-cy="Good"
+                  className={trSelectedLight}
+                >
+                  <td>
+                    <button
```
naziamoff, line 87: I believe it would be better to have two buttons.

naziamoff (changes requested):

naziamoff (approved):
👍

PR #1583, opened 2023-06-12
solution

okosohor:
[DEMO LINK](https://okosohor.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,95 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string,
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state = {
+    selectedGood: '',
+  };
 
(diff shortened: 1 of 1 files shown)


naziamoff (changes requested):
a couple of fixes and we're good

src/App.tsx
```diff
@@ -15,71 +16,94 @@ export const goods = [
...
-);
+              />
+            </>
+          )}
+        </h1>
+
+        <table className="table">
+          <tbody>
+            {goods.map((good) => (
+              <tr
```
33Orange, line 68: Let's create a variable for this condition

```suggestion
            {goods.map((good) => {
              const isSelected = selectedGood === good;
              
              return (<tr
```

33Orange (changes requested):
Almost done

Just a one moment to fix

src/App.tsx
```diff
@@ -15,71 +16,98 @@ export const goods = [
...
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  removeGood = () => (
```
naziamoff, line 28: we don't remove good in this method, we reset selected

naziamoff (approved):
good job! See one suggestion

PR #1603, opened 2023-06-12
solution

vitvadym:
[DEMO LINK](https://vitvadym.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +15,91 @@ export const goods = [
...
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  addGood = (item: string) => {
```
naziamoff, line 27: does it add good to the list? By the name it seems it does
also, let's be more specific, `item` - is not a good name for a variable

src/App.tsx
```diff
@@ -15,71 +15,91 @@ export const goods = [
...
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+  removeGood = () => {
```
naziamoff, line 31: it doesn't remove good from the list

src/App.tsx
```diff
@@ -15,71 +15,91 @@ export const goods = [
...
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+              <button
+                data-cy="ClearButton"
+                type="button"
+                className="delete ml-3"
+                onClick={this.removeGood}
+              />
+            </h1>
+          )
+          : (<h1 className="title">No goods selected</h1>)}
```
naziamoff, line 54:
```suggestion
          )
          : <h1 className="title">No goods selected</h1>}
```

src/App.tsx
```diff
@@ -15,71 +15,91 @@ export const goods = [
...
-          </td>
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
+                key={good}
+                data-cy="Good"
+                className={selectedGood === good
+                  ? 'has-background-success-light'
+                  : ''}
```
naziamoff, line 64: [classnames](https://www.npmjs.com/package/classnames)

naziamoff (changes requested):
not bad, but see a couple of suggestions

PR #1606, opened 2023-06-13
Solution

MaksimKad:
[DEMO LINK](https://MaksimKad.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,86 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string,
+};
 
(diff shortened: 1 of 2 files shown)


naziamoff (changes requested):
good job 👍 
Let's implement some improvements though

src/App.tsx
```diff
@@ -15,71 +16,85 @@ export const goods = [
...
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
```
33Orange, line 60: Let's create a variable for our condition and use it

```suggestion
            {goods.map(good => {
              const isSelected = selectedGood === good;

              return (<tr
```

33Orange (approved):
Well done

PR #1610, opened 2023-06-13
added solution

niaYaro:
[DEMO LINK](https://niaYaro.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
 export const goods = [
   'Dumplings',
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -14,72 +15,94 @@ export const goods = [
...
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+  dismissSelectedGoods = () => {
```
naziamoff, line 34: it selects only one good, so the name of the function lies to me a little

src/App.tsx
```diff
@@ -14,72 +15,94 @@ export const goods = [
...
-);
+              return (
+                <tr
+                  data-cy="Good"
+                  className={isSelected
+                    ? 'has-background-success-light'
+                    : ''}
+                >
+                  <td>
+                    <button
```
naziamoff, line 79: let's create two buttons. It is a good thing that you think of reusability, but readability is our first priority

33Orange (approved):
Well done

PR #1616, opened 2023-06-14
Develop

SofiaLiubchenko:
[DEMO LINK](https://SofiaLiubchenko.github.io/react_goods-selector/)


33Orange (changes requested):
Almost done

A few moments  to fix

And demo link doesn't work
image

PR #1618, opened 2023-06-14
add solution

allagutsul27:
[DEMO LINK](https://allagutsul27.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -58,6 +58,7 @@ export class App extends Component<{}, State> {
 
               return (
                 <tr
+                  key={good}
                   data-cy="Good"
                   className={
                     isSelectedGood


src/App.tsx
```diff
@@ -15,71 +15,84 @@ export const goods = [
...
+                        ? 'button is-info'
+                        : 'button'}
+                      onClick={
+                        isSelectedGood
+                          ? this.removeSelectedGood
+                          : () => this.addSelectedGood(good)
+                      }
+                    >
+                      {isSelectedGood ? '+' : '-'}
+                    </button>
```
etojeDenys, line 84:
```suggestion
        {isSelectedGood ? (
                    <button
                      data-cy='RemoveButton'
                      type="button"
                      className='button is-info'
                      onClick={this.removeSelectedGood}
                    >
                      +
                    </button> ) : (...)
```
since we are changing almost all the props, it is better to write it through the ternary operator, it will be much more readable

etojeDenys (changes requested):

src/App.tsx
```diff
@@ -15,71 +15,87 @@ export const goods = [
...
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
+              return (
+                <tr
+                  data-cy="Good"
```
Viktor-Kost, line 61: add a React `key` to items you map

src/App.tsx
```diff
@@ -15,71 +15,87 @@ export const goods = [
...
+                        onClick={this.removeSelectedGood}
+                      >
+                        +
+                      </button>
+                    ) : (
+                      <button
+                        data-cy="AddButton"
+                        type="button"
+                        className="button"
+                        onClick={() => this.addSelectedGood(good)}
```
Viktor-Kost, line 83: move this handler to an own function

Viktor-Kost (changes requested):
GJ!
Some minor fixes needed

Viktor-Kost (approved):

PR #1620, opened 2023-06-16
solution

mariuszgit:


src/App.tsx
```diff
@@ -15,71 +18,75 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type AppState = {
+  isSelected: string
```
DorotaLeniecDev, line 22: Name this property **selectedGood**, **isSelected** would be suitable if the type of this property is a boolean

src/App.tsx
```diff
@@ -15,71 +18,75 @@ export const goods = [
...
-              -
-            </button>
-          </td>
+        <table className="table">
+          <tbody>
+            {goods.map(good => {
+              const btnClassName = cn({
+                button: true,
+                'is-info': good === isSelected,
+              });
```
DorotaLeniecDev, line 63: If the className should always be added you can write:
```suggestion
              const btnClassName = cn('button', {
                'is-info': good === isSelected,
              });
```

DorotaLeniecDev (changes requested):
Good job, few small comments

DorotaLeniecDev (approved):
Good job!

PR #1622, opened 2023-06-16
Solution

yura-enttsel:
[DEMO LINK](https://yura-enttsel.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,6 @@
-import React from 'react';
+/* eslint-disable jsx-a11y/control-has-associated-label */
+import { Component } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +17,77 @@ export const goods = [
...
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string | null;
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends Component<{}, State> {
+  state: Readonly<State> = {
+    selectedGood: 'Jam',
```
OleziO, line 26: Replace with `goods[goods.indexOf('Jam')] || goods[0]`

OleziO: Its more safe variant

OleziO (changes requested):
image
Minus does not work

mykhalenych (approved):

PR #1625, opened 2023-06-19
Solution

Mar4enkofff:
https://Mar4enkofff.github.io/react_goods-selector/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,8 +1,8 @@
 import React from 'react';
-import 'bulma/css/bulma.css';
 import './App.scss';
+import classNames from 'classnames';
 
-export const goods = [
+const goods: string[] = [
   'Dumplings',
   'Carrot',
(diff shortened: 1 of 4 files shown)


loralevitska (changes requested):
Almost done)

Check class names, try to do your solution according [example](https://mate-academy.github.io/react_goods-selector/)
now the work looks like this
image

src/App.tsx
```diff
@@ -15,71 +16,81 @@ export const goods = [
...
-);
+          )}
+        </h1>
+        <table className="table">
+          <tbody>
+            {goods.map((good) => (
+              <tr
+                data-cy="Good"
+                key={good}
+                className={classNames('', {
```
etojeDenys, line 61:
```suggestion
                className={classNames({
```

etojeDenys (approved):
lgtm

PR #1629, opened 2023-06-20
solution

Kinderwagen:
[DEMO LINK](https://Kinderwagen.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,89 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string,
+};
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +15,89 @@ export const goods = [
...
+                >
+                  <td>
+                    <button
+                      data-cy={
+                        isSelected
+                          ? 'RemoveButton'
+                          : 'AddButton'
+                      }
+                      type="button"
+                      className={`button ${isSelected ? 'is-info' : ''}`}
```
etojeDenys, line 81: do same here

etojeDenys (changes requested):

src/App.tsx
IvanFesenko: Means use classNames library here)

IvanFesenko (approved):

PR #1639, opened 2023-06-25
Develop

Nazarht:
[DEMO LINK](https://nazarht.github.io/react_goods-selector/)


src/App.jsx
```diff
@@ -0,0 +1,156 @@
...
+              Dumplings
+            </td>
+          </tr>
+
+          <tr
+            data-cy="Good"
+            className={food === 'Jam' && 'has-background-success-light'}
+          >
+            <td>
+              {food === 'Jam'
```
OleziO, line 92: Replace with `food === selectedGood`

src/App.jsx
```diff
@@ -0,0 +1,156 @@
...
+            <td data-cy="GoodTitle" className="is-vcentered">
+              Jam
+            </td>
+          </tr>
+
+          <tr
+            data-cy="Good"
+            className={food === 'Garlic' && 'has-background-success-light'}
+          >
+            <td>
```
OleziO, line 124: Use map method to avoid code duplication.

Bad case:
```
<div>...</div>
<div>...</div>
<div>...</div>
<div>...</div>
```

Good case:
```
list.map(item => {
 return  <div>...</div>;
}
```

OleziO (changes requested):
Work version: 
image
Demo: 
image

Add other goods and set `Jam` active by default

PR #1657, opened 2023-06-29
Develop

BipBobBoop:
[DEMO LINK](https://BipBobBoop.github.io/react_goods-selector-js/)


src/App.jsx
```diff
@@ -0,0 +1,74 @@
...
+  'Apple',
+  'Bread',
+  'Fish',
+  'Honey',
+  'Jam',
+  'Garlic',
+];
+
+export const App = () => {
+  const [selectedGood, setValue] = useState('Jam');
```
Anastasiia-Svintsova, line 19: 1) You've already had the 'Jam' in the goods array.
2)  It's recommended to name the setter function based on the specific state variable it's associated with
```suggestion
  const [selectedGood, setSelectedGood] = useState(goods[8]);
```

src/App.jsx
```diff
@@ -0,0 +1,74 @@
...
+
+  return (
+    <main className="section container">
+      {selectedGood
+        ? (
```
Anastasiia-Svintsova, line 24:
```suggestion
      {selectedGood ? (
```

src/App.jsx
```diff
@@ -0,0 +1,74 @@
...
+          <h1 className="title is-flex is-align-items-center">
+            {`${selectedGood} is selected`}
+            <button
+              data-cy="ClearButton"
+              type="button"
+              className="delete ml-3"
+              onClick={() => setValue('')}
```
Anastasiia-Svintsova, line 31: You use this logic for the remove button as well. So, move this logic to a separate function to not repeat yourself

src/App.jsx
```diff
@@ -0,0 +1,74 @@
...
+            />
+          </h1>
+        )
+        : (
+          <h1 className="title is-flex is-align-items-center">
+            No goods selected
+          </h1>
+        )
+    }
```
Anastasiia-Svintsova, line 40:
```suggestion
        ) : (
          <h1 className="title is-flex is-align-items-center">
            No goods selected
          </h1>
        )}
```

src/App.jsx
```diff
@@ -0,0 +1,74 @@
...
+            return (
+              <tr
+                data-cy="Good"
+                className={isSelectedGood && 'has-background-success-light'}
+              >
+                <td>
+                  <button
+                    data-cy={isSelectedGood ? 'RemoveButton' : 'AddButton'}
+                    type="button"
+                    className={isSelectedGood ? 'button is-info' : 'button'}
```
Anastasiia-Svintsova, line 55: In my opinion, it's better to create 2 separate buttons to separate the logic. Let's imagine that your app will grow and you need to use these buttons separately. 
But both approaches can be valid, so it's up to you

Anastasiia-Svintsova (changes requested):
Good start! Let's add a few fixes

src/App.jsx
BipBobBoop: не можу нажаль закинути пул реквест в потрібну директорію... її немає в списку:
image

PR #1667, opened 2023-07-23
Task solution

ArtemBehter:


src/App.tsx
```diff
@@ -15,71 +16,74 @@ export const goods = [
...
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+export const App = () => {
+  const [value, setValue] = useState('Jam');
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+  return (
+    <main className="section container">
+      {value === '' ? (
```
darokrk, line 24: Don't need to check if value is empty, rather use:

`{!value ? (`

src/App.tsx
```diff
@@ -15,71 +16,74 @@ export const goods = [
...
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+          <button
+            data-cy="ClearButton"
+            type="button"
+            className="delete ml-3"
+            onClick={() => {
```
darokrk, line 36: you can extract this logic to external function to be like:

```
  removeSelectedGood = () => {
    setValue('');
  };
```

and just use it in button:

```
<button
     data-cy="ClearButton"
     type="button"
     className="delete ml-3"
     onClick={removeSelectedGood}
/>
```

src/App.tsx
```diff
@@ -15,71 +16,74 @@ export const goods = [
...
+        <tbody>
+          {goods.map(good => (
+            good === value ? (
+              <tr data-cy="Good" className="has-background-success-light">
+                <td>
+                  <button
+                    data-cy="RemoveButton"
+                    type="button"
+                    className="button is-info"
+                    onClick={() => {
```
darokrk, line 53: same as above reuse `removeSelectedGood` function

src/App.tsx
```diff
@@ -15,71 +16,74 @@ export const goods = [
...
+                </td>
+              </tr>
+            ) : (
+              <tr data-cy="Good">
+                <td>
+                  <button
+                    data-cy="AddButton"
+                    type="button"
+                    className="button"
+                    onClick={() => {
```
darokrk, line 72: this can be also extracted to external function e.g :

```
addGood = (good: string) => {
    setValue(good)
};
```

darokrk (changes requested):
Added a few mentions to code 👨🏽‍💻  , also please provide the missing demo link 👀

darokrk (changes requested):
The changes are nice, but what about the missing demo link? Please add a readme file with a demo link to see how the project looks live

darokrk (approved):
Well done 🔥 👏

PR #1673, opened 2023-08-08
add task solution

Luk2asz:
[DEMO LINK](https://Luk2asz.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,3 +1,8 @@
+/* eslint-disable jsx-a11y/control-has-associated-label */
+/* eslint-disable @typescript-eslint/no-shadow */
+/* eslint-disable @typescript-eslint/no-unused-vars */
+/* eslint-disable @typescript-eslint/no-unused-expressions */
+/* eslint-disable no-return-assign */
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +20,124 @@ export const goods = [
...
-        </tr>
-
-        <tr data-cy="Good">
-          <td>
+export class App extends React.Component {
+  state = {
+    selectedGood: 'Jam',
+  };
+
+  reset = () => {
```
darokrk, line 28: hmm your logic looks too much complicated, our clear button click should just clear the state, don't need to manipulate DOM attributes and classes:

```
  clear = () => {
  };
```

you can name it as `clear` or `clearItem` to be more descriptive

src/App.tsx
```diff
@@ -15,71 +20,124 @@ export const goods = [
...
+          button.setAttribute('data-cy', 'AddButton');
+          button.textContent = '+';
+        }
+      }
+    });
+
+    this.setState({ selectedGood: null });
+  };
+
+  handleClick = (good: string) => {
```
darokrk, line 47: this logic looks even more complicated as removing item, this logic should only update selectedGood in state, not doing any manipulation on selectors or loops, they are redundant, try this one:

```
 selectedItem = (good: string) => {
    this.setState({ selectedGood: good });
 };
```

src/App.tsx
```diff
@@ -15,71 +20,124 @@ export const goods = [
...
+    }
+  };
+
+  render() {
+    const { selectedGood } = this.state;
+
+    return (
+      <main className="section container">
+        {selectedGood ? (
+          <h1 className="title is-flex is-align-items-center">
```
darokrk, line 93: here you can try use `classNames()` function to handle conditionally change css:

```
<h1
  className={classNames(
    'title',
     {
       'is-flex is-align-items-center': selectedGood,
      },
   )}
 >
```

remember to install [https://www.npmjs.com/package/classnames](url) and import this module `import classNames from 'classnames';`

src/App.tsx
```diff
@@ -15,71 +20,124 @@ export const goods = [
...
+            {selectedGood}
```
darokrk, line 94: then here you dont need to create double `<h1>` elements, just handle text inside, e.g. :

```
 {selectedGood
            ? `${selectedGood} is selected`
            : 'No goods selected'}
```

src/App.tsx
```diff
@@ -15,71 +20,124 @@ export const goods = [
...
-      </tbody>
-    </table>
-  </main>
-);
+              className="delete ml-3"
+              onClick={this.reset}
+            />
+          </h1>
+        ) : (
+          <h1 className="title">
```
darokrk, line 106: this element and check is redundant, please remove it

src/App.tsx
```diff
@@ -15,71 +20,124 @@ export const goods = [
...
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Garlic
-          </td>
-        </tr>
-      </tbody>
-    </table>
-  </main>
-);
+              className="delete ml-3"
+              onClick={this.reset}
```
darokrk, line 102: now here we will change it to be:

`onClick={this.clear}` or `onClick={this.clearItem}` depending what name you will chose

src/App.tsx
```diff
@@ -15,71 +20,124 @@ export const goods = [
...
+        )}
+
+        <table className="table">
+          <tbody>
+            {goods.map(good => (
+              <tr
+                data-cy="Good"
+                id={good}
+                key={good}
+                className={good === 'Jam'
```
darokrk, line 118: now here we ale can use our earlier imported classNames() function to conditionaly display css classes:

```
<tr
   data-cy="Good"
   className={classNames(
   { 'has-background-success-light': good === selectedGood },
  )}
>
```

src/App.tsx
```diff
@@ -15,71 +20,124 @@ export const goods = [
...
+                  ? 'has-background-success-light'
+                  : ''}
+              >
+                <td>
+                  <button
+                    data-cy={good === 'Jam' ? 'RemoveButton' : 'AddButton'}
```
darokrk, line 124: here don't need to compare the hardcoded value 'Jam' instead we should compare the iterated item to the current selectedGood state to avoid errors:

```
 data-cy={good === selectedGood
                      ? 'RemoveButton'
                      : 'AddButton'}
```

src/App.tsx
```diff
@@ -15,71 +20,124 @@ export const goods = [
...
+                    type="button"
+                    className={good === 'Jam' ? 'button is-info' : 'button'}
```
darokrk, line 126: here either we can use our classNames() module function to conditional css switch:

```
 className={classNames(
   'button',
    { 'is-info': good === selectedGood },
 )}
```

src/App.tsx
```diff
@@ -15,71 +20,124 @@ export const goods = [
...
+                    onClick={() => this.handleClick(good)}
```
darokrk, line 127: Here we will just update our state based on selected item clicked:

```
onClick={() => this.selectedItem(
  selectedGood === good ? '' : good,
)}
```

src/App.tsx
```diff
@@ -15,71 +20,124 @@ export const goods = [
...
+                  >
+                    {good === 'Jam' ? '-' : '+'}
```
darokrk, line 129: here like above we should compare our selected good directly from state rather then hardcoded 'Jam' value:

`{good === selectedGood ? '-' : '+'}`

darokrk (changes requested):
The code need some improvements 🛠️   please follow my mentions and try to fix it 👨🏽‍💻

darokrk (approved):
Perfect 👏🏽  easier and cleaner 🔥 🥇 !

PR #1685, opened 2023-08-28
just testing

cardinal1991:


choeqq (changes requested):
We should be able to select diffent item from the list without having to click the `-` on currently selected item, also your styling is little bit off 😄

choeqq (changes requested):
Looks better, but we should also be able to unselect the item with `-` on selected item, it should work the same as this button:
image

choeqq (approved):

PR #1686, opened 2023-08-29
Solution

Vira-V:
[DEMO LINK](https://Vira-V.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,89 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +15,89 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: null | string;
```
Sancz0pansa, line 19: Shouldn't it be just string?

Sancz0pansa (changes requested):
Demo link is not working.

choeqq (approved):

PR #1709, opened 2023-09-17
Solution

KarolinaWiktoria:
[DEMO LINK](https://KarolinaWiktoria.github.io/react_goods-selector/)


src/App.tsx
```diff
@@ -15,71 +19,80 @@ export const goods = [
...
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
-
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, AppState> {
+  state: Readonly<AppState> = {
+    selectedGood: 'Jam',
```
OleziO, line 24: Replace with `goods[goods.indexOf('Jam')] || goods[0]`
Its more safe variant

src/App.tsx
```diff
@@ -15,71 +19,80 @@ export const goods = [
...
 
-        <tr data-cy="Good" className="has-background-success-light">
-          <td>
+    return (
+      <main className="section container">
+        {selectedGood ? (
+          <h1 className="title is-flex is-align-items-center">
+            {selectedGood}
+            {' '}
+            is selected
```
OleziO, line 40: Use template strings.

[More info.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

OleziO (changes requested):

src/App.tsx
KarolinaWiktoria: I had to use {/* eslint-disable-next-line react/jsx-one-expression-per-line */}, because of linter warnings.

etojeDenys (approved):
good job

PR #1713, opened 2023-11-20
add task solution

netzelnatalia:
[DEMO LINK](https://netzelnatalia.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,101 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string,
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state: Readonly<State> = {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +15,101 @@ export const goods = [
...
+                    key={good}
+                  >
+                    <td>
+                      <button
+                        data-cy="RemoveButton"
+                        type="button"
+                        className="button is-info"
+                        onClick={() => {
+                          this.setState({ selectedGood: '' });
+                        }}
```
volodymyr-soltys97, line 69: Consider will be better if you create a helper function for this logic and use it everywhere

src/App.tsx
```diff
@@ -15,71 +15,101 @@ export const goods = [
...
+                      >
```
volodymyr-soltys97, line 70: Now the essentially identical content is rendered here in two blocks, it is unnecessary, try to specify a certain class by condition, you can use the classnames library to specify classes

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

nazarmatsevych (approved):
Great work!

PR #1718, opened 2023-12-04
Solution

zaura333:
[DEMO LINK](https://zaura333.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -24,6 +24,14 @@ type AppState = {
 export class App extends Component<{}, AppState> {
   state = { selectedGood: 'Jam' }
 
+  handleClearClick = () => {
+    this.setState({ selectedGood: '' });
+  }
+
+  handleListClick = (good: string) => {
+    this.setState({ selectedGood: good });
+  }
+
   render() {
     const { selectedGood } = this.state;
 
@@ -39,7 +47,7 @@ export class App extends Component<{}, AppState> {
                 data-cy="ClearButton"
                 type="button"
                 className="delete ml-3"
-                onClick={() => this.setState({ selectedGood: '' })}
+                onClick={this.handleClearClick}
               />
             </h1>
           )}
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +17,64 @@ export const goods = [
...
+          ? <h1 className="title">No goods selected</h1>
+          : (
+            <h1 className="title is-flex is-align-items-center">
+              {`${selectedGood} is selected`}
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+              <button
+                data-cy="ClearButton"
+                type="button"
+                className="delete ml-3"
+                onClick={() => this.setState({ selectedGood: '' })}
```
jacekkosiba, line 42: If I understand the task correctly, we should use dedicated methods instead of inline handlers

from checklist.md:
> 2. [CODE KNOWLEDGE] - Create separate methods instead of creating inline event handlers

src/App.tsx
```diff
@@ -15,71 +17,64 @@ export const goods = [
...
+              />
+            </h1>
+          )}
+        <table className="table">
+          <tbody>
+            {goods.map((good) => {
+              const isSelected = good === selectedGood;
+              const buttonClass = cn({ button: true, 'is-info': isSelected });
```
jacekkosiba, line 50: Took me a while to get whats going on with the 'cn' :D That's a nice feature!

src/App.tsx
```diff
@@ -15,71 +17,64 @@ export const goods = [
...
-          </td>
-        </tr>
-      </tbody>
-    </table>
-  </main>
-);
+              return (
+                <tr data-cy="Good" className={trClass} key={good}>
+                  <td>
+                    <button
```
jacekkosiba, line 58: This solution is really cool and it seems more complex and ambitious than just rendering a different button based on whether an item is or is not a selected item, although in my opinion it can get a bit hard to read. Also, it forces you to use 3 ternary operators instead of one, so I'm not sure if it's the most optimal solution

A simpler approach would be
`good !== selectedGood
      ? // add button
      : // clear button`

jacekkosiba (approved):
Great and ambitious solution!

mbruhler (approved):

PR #1720, opened 2023-12-04
task solved

TommasoGiacopini:
[DEMO LINK](https://TommasoGiacopini.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,86 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+interface State {
+  goodSelected: string | null;
+}
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state: State = {
+    goodSelected: 'Jam',
+  };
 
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -15,71 +15,86 @@ export const goods = [
...
+                      className="button"
+                      onClick={() => this.onSelect(good)}
+                    >
+                      +
+                    </button>
+                  </td>
+                  <td data-cy="GoodTitle" className="is-vcentered">
+                    {good}
+                  </td>
+                </tr>
```
witflash, line 92: You can use a slightly different approach:
```
{goods.map(good => (
  <tr
    data-cy="Good"
    key={good}
    className={goodSelected === good ? "has-background-success-light" : ""}
  >
    <td>
      {goodSelected === good ? (
        <button
          // all props for the selected good button
        />
      ) : (
        <button
          // all props for the add button
        />
      )}
    </td>
  </tr>
```
So, your code will be shorter.

witflash (approved):
Good work!

I left just one comment to think about your code optimization.
But it's not so critical. Your code may be an alternative solution.

src/App.tsx
TommasoGiacopini: Yes, that makes sense, thank you!

PR #1725, opened 2023-12-05
solution

bertsu02:
[DEMO LINK](https://bertsu02.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,71 +15,77 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGoods: string | null;
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
+  state: State = {
+    selectedGoods: 'Jam',
+  };
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  render() {
+    const { selectedGoods } = this.state;
 
-    <table className="table">
-      <tbody>
-        <tr data-cy="Good">
-          <td>
-            <button
-              data-cy="AddButton"
-              type="button"
-              className="button"
-            >
-              +
-            </button>
-          </td>
-
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
-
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +15,77 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGoods: string | null;
```
jacekkosiba, line 19: This is 100% correct, jus wanted to point out that it also could be
`selectedGoods ?: string,`

src/App.tsx
```diff
@@ -15,71 +15,77 @@ export const goods = [
...
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+export class App extends React.Component<{}, State> {
```
jacekkosiba, line 22: Again, perfectly well written, but could be shorten a bit with
`class App extends Component`

src/App.tsx
```diff
@@ -15,71 +15,77 @@ export const goods = [
...
+        {selectedGoods
+          ? (
+            <h1 className="title is-flex is-align-items-center">
+              {`${selectedGoods} is selected`}
+              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+              <button
+                data-cy="ClearButton"
+                type="button"
+                className="delete m1-3"
+                onClick={() => this.setState({ selectedGoods: '' })}
```
jacekkosiba, line 41: adding a dedicated method(s) for handling selecting and clearing our items would probably be a bit more optimal

jacekkosiba (approved):
Perfect! I just have some minor improvement suggestions.

src/App.tsx
mbruhler, line 41: use custom handlers

src/App.tsx
```diff
@@ -15,71 +15,77 @@ export const goods = [
...
+                          data-cy="AddButton"
+                          type="button"
+                          className="button"
+                          onClick={() => this.setState({ selectedGoods: good })}
+                        >
+                          +
+                        </button>
+                      )
+                      : (
+                        <button
```
mbruhler, line 70: image
your button is too close to the name, it is not like that originally

mbruhler (changes requested):

src/App.tsx
mvjl000: @jacekkosiba not quite the same.
Keep in mind that `selectedGoods?: string` is equal to `selectedGoods: string | undefined`, not `selectedGoods: string | null`

mvjl000 (approved):
All good!

PR #1729, opened 2023-12-06
Solution

web3-sorcerer:
[DEMO LINK](https://web3-sorcerer.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { Component } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +15,67 @@ export const goods = [
...
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+  handleItemClick = (itemName: string) => {
+    if (itemName === this.state.selected) {
+      this.setState({ selected: '' });
+    } else {
+      this.setState({ selected: itemName });
+    }
+  };
```
JoyCoffeeAddict, line 29: look how Miłosz sugested it: 
```ts
handleItemClick = (itemName: string) => {
  if (itemName === this.state.selected) {
    this.setState({ selected: '' });
    return;
  };
  this.setState({ selected: itemName });
}
```

You should avoid `else-ing` when possible, and just add a simple early return

mvjl000 (approved):
Well done!

PR #1735, opened 2024-02-09
Solution

wsadocha:
[DEMO LINK](https://wsadocha.github.io/react_goods-selector/)


BogdanMaliuta (approved):
Good job!
But I don`t understand why did you remove func component and used class.In the future, you'll only use functional components, also better way to condition styles it`s use library classNames

PR #1736, opened 2024-02-12
Solution

Krzychu92:
https://krzychu92.github.io/react_goods-selector/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -38,7 +38,7 @@ export class App extends React.Component {
         {this.state.selectedGood
           ? (
             <h1 className="title is-flex is-align-items-center">
-              {`${this.state.selectedGood} is selected`}
+              {`${selectedGood} is selected`}
               {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
               <button
                 onClick={this.handleClear}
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +15,85 @@ export const goods = [
...
+            {goods.map((good) => {
+              return (
+                <tr
+                  key={good}
+                  data-cy="Good"
+                  className={good === selectedGood
+                    ? 'has-background-success-light'
+                    : undefined}
+                >
+                  {good === this.state.selectedGood
```
yevhenii-pyl, line 63: Since there are no button components for different scenarios,  I'd rather modify a button properties instead of rendering different buttons. 
```
                        <button
                          onClick={good === this.state.selectedGood ? this.handleClear : ....}
                          data-cy={good === this.state.selectedGood ? ....}
                          type="button"
                          className=`button ${good === this.state.selectedGood && ....}`
                        >
                          {good === this.state.selectedGood ? ...}
                        </button>
```
while this approach may be arguable, there is definitely no need to render `<td>` twice, you can place your render condition inside.

yevhenii-pyl, line 63: You have it extracted right after `render() {`  => so there is no need to type `this.state.selectedGood`, you can just go with `selectedGood`. 
This is a solid approach, stick to it :)

src/App.tsx
```diff
@@ -15,71 +15,85 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+export class App extends React.Component {
```
yevhenii-pyl, line 18: I see that lot's of students are refactoring functional component into a class component in this taks. 
I don't see it as a mistake, but more out of a curiosity question: why not to use FC?

yevhenii-pyl (changes requested):
Good job! Let's just make a few small adjustments .

yevhenii-pyl (approved):
Nice job!
So why using class component anyways? Don't hesitate to answer under my comment or at DM on a platform :)

Krzychu92:
> Nice job! So why using class component anyways? Don't hesitate to answer under my comment or at DM on a platform :)

maybe becouse it was just after a videos about a class components and before I have learned about a hooks. Now I would probably use a function component with a hook to solve that. Is there any way to do it as a functional component without hook?

PR #1737, opened 2024-02-13
solution

Patryk-Buczkowski:
[DEMO LINK](https://Patryk-Buczkowski.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -22,7 +22,11 @@ export class App extends React.Component {
   };
 
   handlerSelect = (good: string) => {
-    this.setState({ selectedGood: good });
+    if (this.state.selectedGood === good) {
+      return this.handlerClear;
+    }
+
+    return this.setState({ selectedGood: good });
   };
 
   handlerClear = () => {
@@ -30,14 +34,6 @@ export class App extends React.Component {
       selectedGood: '',
     });
   };
-
-  helperSelector = (good: string): (() => void) | undefined => {
-    if (this.state.selectedGood === good) {
-      return this.handlerClear;
-    }
-
-    return () => this.handlerSelect(good);
-  }
 
   render() {
     const { selectedGood } = this.state;
@@ -63,6 +59,7 @@ export class App extends React.Component {
           <tbody>
             {goods.map((good: string) => (
               <tr
+                key={good}
                 data-cy="Good"
                 className={cn({
                   'has-background-success-light': selectedGood === good,
@@ -70,19 +67,16 @@ export class App extends React.Component {
               >
                 <td>
                   <button
-                    onClick={this.helperSelector(good)}
-                    data-cy={selectedGood === good
-                      ? 'RemoveButton'
-                      : 'AddButton'}
+                    onClick={() => this.handlerSelect(good)}
+                    data-cy={
+                      selectedGood === good ? 'RemoveButton' : 'AddButton'
+                    }
                     type="button"
-                    className={cn({
-                      button: true,
-                      'button is-info': selectedGood === good,
+                    className={cn('button', {
+                      'is-info': selectedGood === good,
                     })}
                   >
-                    {selectedGood === good
-                      ? '-'
-                      : '+'}
+                    {selectedGood === good ? '-' : '+'}
                   </button>
                 </td>
 
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
+import cn from 'classnames';
 
 export const goods = [
   'Dumplings',
@@ -21,9 +22,7 @@ export class App extends React.Component {
   };
 
   handlerSelect = (good: string) => {
-    return () => {
-      this.setState({ selectedGood: good });
-    };
+    this.setState({ selectedGood: good });
   };
 
   handlerClear = () => {
@@ -32,12 +31,20 @@ export class App extends React.Component {
     });
   };
 
+  helperSelector = (good: string): (() => void) | undefined => {
+    if (this.state.selectedGood === good) {
+      return this.handlerClear;
+    }
+
+    return () => this.handlerSelect(good);
+  }
+
   render() {
     const { selectedGood } = this.state;
 
     return (
       <main className="section container">
-        {selectedGood === '' ? (
+        {!selectedGood ? (
           <h1 className="title">No goods selected</h1>
         ) : (
           <h1 className="title is-flex is-align-items-center">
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -15,71 +15,105 @@ export const goods = [
...
+                      <button
+                        onClick={this.handlerSelect(good)}
+                        data-cy="AddButton"
+                        type="button"
+                        className="button"
+                      >
+                        +
+                      </button>
+                    )}
+                  </td>
```
BogdanMaliuta, line 90: for me better option for this conditional rendering it`s just change your "+" and "-", not buttons, just make handlerSelect better, add condition(check selectedGood and good what user wants to select)

src/App.tsx
```diff
@@ -15,71 +15,105 @@ export const goods = [
...
-    <h1 className="title">No goods selected</h1>
+export class App extends React.Component {
+  state = {
+    selectedGood: 'Jam',
+  };
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
+  handlerSelect = (good: string) => {
+    return () => {
```
BogdanMaliuta, line 24: For what are you using return

src/App.tsx
```diff
@@ -15,71 +15,105 @@ export const goods = [
...
 
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button
-        data-cy="ClearButton"
-        type="button"
-        className="delete ml-3"
-      />
-    </h1>
+  handlerClear = (event: { stopPropagation: () => void; }) => {
+    event.stopPropagation();
```
BogdanMaliuta, line 30: Why did you use it?

BogdanMaliuta (changes requested):

BogdanMaliuta (changes requested):
Hello, nice job, but try to improve your code.
Make your handlerSelect more reusable(you realy dont need to create handlerClear), add condition and remove return and event.stopPropagation();.

src/App.tsx
```diff
@@ -38,79 +37,56 @@ export class App extends React.Component {
...
-              {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-              <button
-                onClick={this.handlerClear}
-                data-cy="ClearButton"
-                type="button"
-                className="delete ml-3"
-              />
-            </h1>
-          )}
+        {selectedGood === '' ? (
```
yevhenii-pyl, line 40: We can simplify this condition by checking `!!selectedGood`. It will just look cleaner.

src/App.tsx
```diff
@@ -38,79 +37,56 @@ export class App extends React.Component {
```
yevhenii-pyl, line 37: As a rule of thumb, you can destructure all required properties from the state here. 

`{ selectedGood } = this.state` to avoid using this.state.... each time you need some data from state at your JSX.

src/App.tsx
yevhenii-pyl: I believe Bogdan wanted to emphasise that there is no need to return something from that function, you can have it like that: 
```
  handlerSelect = (good: string) => {
      this.setState({ selectedGood: good });
  };
  ```

File: src/App.tsx [hunk 6]
```diff
@@ -15,71 +15,81 @@ export const goods = [
...
 
-        <tr data-cy="Good" className="has-background-success-light">
-          <td>
+    return (
+      <main className="section container">
+        {selectedGood === '' ? (
+          <h1 className="title">No goods selected</h1>
+        ) : (
+          <h1 className="title is-flex is-align-items-center">
+            {`${this.state.selectedGood} is selected`}
```
yevhenii-pyl, line 44: You have destructured it from the state at line 36 :)

yevhenii-pyl (changes requested):
Nice job!  🚀 

Let's just add a bit of a finishing touch and we are good to go.

src/App.tsx
BogdanMaliuta: still not fixed

src/App.tsx
BogdanMaliuta: same here

src/App.tsx
```diff
@@ -15,71 +15,81 @@ export const goods = [
...
+        <table className="table">
+          <tbody>
+            {goods.map((good: string) => (
+              <tr
+                data-cy="Good"
+                className={
+                  selectedGood === good
+                    ? 'has-background-success-light'
+                    : undefined
+                }
```
BogdanMaliuta, line 64: you can use it in another way, without undefined for exm:
cn={
  'style-selector': condition
}

src/App.tsx
```diff
@@ -15,71 +15,81 @@ export const goods = [
...
+              >
+                <td>
+                  <button
+                    onClick={selectedGood === good
+                      ? this.handlerClear
+                      : this.handlerSelect(good)}
```
BogdanMaliuta, line 70: pass into onClick only function not a result of function(this.handlerSelect(good)),
if you want to pass args into your func you can return result of this function from callback, exm: 
() => {yourFunc(arg)}

src/App.tsx
```diff
@@ -15,71 +15,81 @@ export const goods = [
...
+                    data-cy={selectedGood === good
+                      ? 'RemoveButton'
+                      : 'AddButton'}
+                    type="button"
+                    className={selectedGood === good
+                      ? 'button is-info'
+                      : 'button'}
```
BogdanMaliuta, line 77: think about how to not duplicate your button selector

BogdanMaliuta (changes requested):
Looks good, but still need some fixes

src/App.tsx
```diff
@@ -15,71 +16,84 @@ export const goods = [
...
-            <button
-              data-cy="RemoveButton"
-              type="button"
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+    return () => this.handlerSelect(good);
+  }
```
BogdanMaliuta, line 40: don`t realy need to create separate function for check is this good selected, just add this to your handlerSelect

src/App.tsx
```diff
@@ -15,71 +16,84 @@ export const goods = [
...
+              >
+                <td>
+                  <button
+                    onClick={this.helperSelector(good)}
+                    data-cy={selectedGood === good
+                      ? 'RemoveButton'
+                      : 'AddButton'}
+                    type="button"
+                    className={cn({
+                      button: true,
```
BogdanMaliuta, line 79:
```suggestion
                      button,
```

src/App.tsx
```diff
@@ -15,71 +16,84 @@ export const goods = [
...
+                      'button is-info': selectedGood === good,
```
BogdanMaliuta, line 80:
```suggestion
                      'is-info': selectedGood === good,
```

src/App.tsx
```diff
@@ -15,71 +16,84 @@ export const goods = [
...
+            {goods.map((good: string) => (
+              <tr
+                data-cy="Good"
+                className={cn({
+                  'has-background-success-light': selectedGood === good,
+                })}
+              >
+                <td>
+                  <button
+                    onClick={this.helperSelector(good)}
```
BogdanMaliuta, line 73: in onClick atribute you need to pass function, not a result of function its can be callBack, if you need to pass some args, 
exm.: {() => yourFunc(args)}

BogdanMaliuta (approved):
Nice try, but in fix please all things described in my comments

PR #1749, opened 2024-03-28
React goods selector

itsXenthis:
[DEMO LINK](https://itsXenthis.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import classNames from 'classnames';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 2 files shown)


pawelkloc (approved):
To me it seems ok :)

src/App.tsx
```diff
@@ -15,58 +16,67 @@ export const goods = [
...
+      ) : (
+        <h1 className="title is-flex is-align-items-center">
+          {`${good} is selected`}
+          <button
+            data-cy="ClearButton"
+            type="button"
+            className="delete ml-3"
+            onClick={() => {
+              SetGood('');
+            }}
```
nataliaklonowska, line 35: Create separate methods instead of creating inline event handlers (if we do so - we don't create new arrow function after component re-rendering) [[checklist]](https://github.com/mate-academy/react_goods-selector/blob/master/checklist.md)

src/App.tsx
```diff
@@ -15,58 +16,67 @@ export const goods = [
...
+                    className={
+                      good === initialGood ? 'button is-info' : 'button'
+                    }
+                    onClick={() => {
+                      if (good === initialGood) {
+                        SetGood('');
+                      } else {
+                        SetGood(initialGood);
+                      }
+                    }}
```
nataliaklonowska, line 67: Create separate methods instead of creating inline event handlers (if we do so - we don't create new arrow function after component re-rendering) [[checklist]](https://github.com/mate-academy/react_goods-selector/blob/master/checklist.md)

nataliaklonowska (changes requested):

nataliaklonowska (approved):

PR #1754, opened 2024-03-28
solution

etuzimek:


bartKarb (approved):
Well done!
Disclaimer: I saw, you added   

`"jsx":` "react",`

Into isconfig.json

I was doing the same to repair linter, however I discovered, that instead of adding this formula, all we need to do is to type 'npm i' after cloning repository, then 'npm start', then close VS Code, and finally open your repository with VS Code again :)

src/App.tsx
```diff
@@ -15,58 +15,73 @@ export const goods = [
...
-              data-cy="RemoveButton"
+              data-cy="ClearButton"
               type="button"
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+              className="delete ml-3"
+              onClick={() => this.setState({ selectedGood: '' })}
```
nataliaklonowska, line 36: Create separate methods instead of creating inline event handlers (if we do so - we don't create new arrow function after component re-rendering) [[checklist]](https://github.com/mate-academy/react_goods-selector/blob/master/checklist.md)

nataliaklonowska (changes requested):

src/App.tsx
etuzimek: separate methods added

etuzimek:
> I was doing the same to repair linter, however I discovered, that instead of adding this formula, all we need to do is to type 'npm i' after cloning repository, then 'npm start', then close VS Code, and finally open your repository with VS Code again :)

Thanks for info, will check in next repo :)

choeqq (approved):

PR #1759, opened 2024-04-15
Solution

przwojwwp:
[DEMO LINK](https://przwojwwp.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -63,23 +63,15 @@ export class App extends React.Component {
                 }
               >
                 <td>
-                  {selectedGood !== good ? (
-                    <button
-                      data-cy="AddButton"
-                      type="button"
-                      className="button"
-                    >
-                      +
-                    </button>
-                  ) : (
-                    <button
-                      data-cy="RemoveButton"
-                      type="button"
-                      className="button is-info"
-                    >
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -15,58 +15,81 @@ export const goods = [
...
+              <tr
+                data-cy="Good"
+                key={index}
+                onClick={() => this.handleClick(good)}
+                className={
+                  selectedGood === good ? 'has-background-success-light' : ''
+                }
+              >
+                <td>
+                  {selectedGood !== good ? (
```
yevhenii-pyl, line 66: Instead of rendering two buttons here, we should use one html element passing props conditionally: 

```
<button
  data-cy={selectedGood === good ?"RemoveButton" :"AddButton"}
  ...
>
{selectedGood === good .... }
</button>
```

This approach can be easily evolved further by creating a reusable(!) button component (f.e. ClassicButton) that receives prop.

yevhenii-pyl (changes requested):
Good job 💪 
Just one minor adjustment with button and we are good to go.

yevhenii-pyl (approved):
Nice! 🥇

PR #1761, opened 2024-05-14
Solution

aMtiIV:
[DEMO LINK](https://aMtiIV.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,58 +15,83 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+type State = {
+  selectedGood: string;
+};
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button data-cy="ClearButton" type="button" className="delete ml-3" />
-    </h1>
+export class App extends React.Component<{}, State> {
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -15,58 +15,83 @@ export const goods = [
...
-          </td>
-        </tr>
+  handleAdd = (good: string) => {
+    this.setState({ selectedGood: good });
+  };
 
-        <tr data-cy="Good" className="has-background-success-light">
-          <td>
+  render(): React.ReactNode {
+    const { selectedGood } = this.state;
```
yevhenii-pyl, line 36: 👍

src/App.tsx
```diff
@@ -15,58 +15,83 @@ export const goods = [
...
+
+    return (
+      <main className="section container">
+        {selectedGood?.length ? (
```
yevhenii-pyl, line 40: Since selectedGood is a string - there is no need to check for .length, will be quite enough to validate it like `!!selectedGood` (or, even as simple as just `selectedGood`).

src/App.tsx
```diff
@@ -15,58 +15,82 @@ export const goods = [
...
+              return (
+                <tr
+                  data-cy="Good"
+                  key={good}
+                  className={
+                    isGoodSelected ? 'has-background-success-light' : ''
+                  }
+                >
+                  <td>
+                    {
```
yevhenii-pyl, line 69: unnecessary `{` here and `}` at the end

yevhenii-pyl (approved):
Good job! 👍

PR #1763, opened 2024-05-21
task solution

meljaszuk:
[DEMO LINK](https://meljaszuk.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -15,58 +15,73 @@ export const goods = [
...
-            </button>
-          </td>
+      <table className="table">
+        <tbody>
+          {goods.map(good => (
+            <tr
+              data-cy="Good"
+              key={good}
+              className={
+                selectedGood === good ? 'has-background-success-light' : ''
```
BogdanMaliuta, line 73: you dont need to pass empty className and use ternary, 
you can do this like that, 'className': condition

src/App.tsx
```diff
@@ -15,58 +15,73 @@ export const goods = [
...
+type Good = (typeof goods)[number];
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button data-cy="ClearButton" type="button" className="delete ml-3" />
-    </h1>
+export const App: React.FC = () => {
+  const [selectedGood, setSelectedGoods] = useState<Good | null>('Jam');
+  const selectedHandler = (good: Good | null) => {
```
BogdanMaliuta, line 22:
```suggestion
  const selectedGoodHandler = (good: Good | null) => {
```
better to understand

BogdanMaliuta (approved):
Nice work!
Next time try to use CN library in another way

PR #1765, opened 2024-07-11
My solution

Karanelus:
[DEMO LINK](https://<your_account>.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import './App.scss';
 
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -15,58 +15,74 @@ export const goods = [
...
-    <h1 className="title">No goods selected</h1>
+export const App: React.FC = () => {
+  const [choosenProduct, setChoosenProduct] = useState<string>(`Jam`);
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button data-cy="ClearButton" type="button" className="delete ml-3" />
-    </h1>
+  const onClickDeselect = () => {
```
Zibi95, line 21: We call this function handlers so most often names of those function starts with `handleClick` .etc

src/App.tsx
```diff
@@ -15,58 +15,74 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+export const App: React.FC = () => {
+  const [choosenProduct, setChoosenProduct] = useState<string>(`Jam`);
```
Zibi95, line 19: If you add default value then typescript will infer the type of state without problems.

src/App.tsx
```diff
@@ -15,58 +15,74 @@ export const goods = [
...
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+            return (
+              <tr
+                data-cy="Good"
+                key={i}
+                className={isGoodChoosed ? 'has-background-success-light' : ''}
```
Zibi95, line 54: In every React project from Mate, there is a default package installed called classnames that allows you to write class name condition logic with better syntax.

src/App.tsx
```diff
@@ -15,58 +15,74 @@ export const goods = [
...
-              type="button"
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+            return (
+              <tr
+                data-cy="Good"
+                key={i}
```
Zibi95, line 53: In this case index as key won't cause any problems but its better not to use it at all. Always look for unique value from the values that you map.

Zibi95 (approved):
Good job! Check below comments for future reference. 💯

PR #1766, opened 2024-07-13
solution

Kozubowicz:
[DEMO LINK](https://Kozubowicz.github.io/react_goods-selector/)


src/App.tsx
```diff
@@ -15,58 +16,67 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+export const App: React.FC = () => {
+  const [selectedGood, setSelectedGood] = useState<string>('Jam');
```
natalia-klonowska, line 20: <string> is unnecessary because TypeScript can automatically infer type from initial value ;)

src/App.tsx
```diff
@@ -15,58 +16,67 @@ export const goods = [
...
-          </td>
+      {selectedGood && (
+        <h1 className="title is-flex is-align-items-center">
+          {selectedGood} is selected
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+          <button
+            data-cy="ClearButton"
+            type="button"
+            className="delete ml-3"
+            onClick={() => setSelectedGood('')}
```
natalia-klonowska, line 34: [checklist](https://github.com/mate-academy/react_goods-selector/blob/master/checklist.md): Create separate methods instead of creating inline event handlers (if we do so - we don't create new arrow function after component re-rendering)

natalia-klonowska (changes requested):

PR #1769, opened 2024-09-16
Develop

przwojwwp:
[DEMO LINK](https://przwojwwp.github.io/react_goods-selector/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,58 +15,99 @@ export const goods = [
   'Garlic',
 ];
 
-export const App: React.FC = () => (
-  <main className="section container">
-    <h1 className="title">No goods selected</h1>
+export class App extends React.Component {
+  state = {
+    selectedGood: 'Jam',
+  };
 
-    <h1 className="title is-flex is-align-items-center">
-      Jam is selected
-      {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
-      <button data-cy="ClearButton" type="button" className="delete ml-3" />
-    </h1>
+  render() {
+    const { selectedGood } = this.state;
 
-    <table className="table">
-      <tbody>
-        <tr data-cy="Good">
-          <td>
-            <button data-cy="AddButton" type="button" className="button">
-              +
-            </button>
-          </td>
+    return (
+      <main className="section container">
+        <h1 className="title">No goods selected</h1>
 
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Dumplings
-          </td>
-        </tr>
-
-        <tr data-cy="Good" className="has-background-success-light">
-          <td>
+        <h1 className="title is-flex is-align-items-center">
+          {selectedGood} is selected
+          {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+          {selectedGood !== '' && (
             <button
-              data-cy="RemoveButton"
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -15,58 +15,83 @@ export const goods = [
...
-              data-cy="RemoveButton"
-              type="button"
-              className="button is-info"
-            >
-              -
-            </button>
-          </td>
+    return (
+      <main className="section container">
+        {selectedGood === '' ? (
```
BogdanMaliuta:
```suggestion
        {!selectedGood ? (
```

src/App.tsx
```diff
@@ -15,58 +15,83 @@ export const goods = [
...
+        ) : (
+          <h1 className="title is-flex is-align-items-center">
+            {selectedGood} is selected
+            {/* eslint-disable-next-line jsx-a11y/control-has-associated-label */}
+            {selectedGood !== '' && (
+              <button
+                data-cy="ClearButton"
+                type="button"
+                className="delete ml-3"
+                onClick={() => this.setState({ selectedGood: '' })}
```
BogdanMaliuta: you have handler  handleRemoveGood

src/App.tsx
```diff
@@ -15,58 +15,83 @@ export const goods = [
...
-        </tr>
+        <table className="table">
+          <tbody>
+            {goods.map((good, index) => {
+              return (
+                <tr
+                  key={index}
+                  data-cy="Good"
+                  className={
+                    selectedGood === good ? 'has-background-success-light' : ''
```
BogdanMaliuta, line 64:
```suggestion
                     'has-background-success-light' :  selectedGood === good
```

src/App.tsx
```diff
@@ -15,58 +15,83 @@ export const goods = [
...
+                      <button
+                        data-cy="AddButton"
+                        type="button"
+                        className="button"
+                        onClick={() => this.handleSelectGood(good)}
+                      >
+                        +
+                      </button>
+                    )}
+                  </td>
```
BogdanMaliuta: just change your content(+/-) using condition, and improve your handleRemoveGood by checking inside is selected good === good, and clear if true

BogdanMaliuta (changes requested):
Nice, but your code needs some changes

src/App.tsx
przwojwwp: className={
                    selectedGood === good && 'has-background-success-light'
                  }
                  
                  Type 'false | "has-background-success-light"' is not assignable to type 'string | undefined'.

zibi-dev: Yea i guess that this syntax requires `classnames` package

src/App.tsx
```diff
@@ -15,58 +15,88 @@ export const goods = [
...
-          <td data-cy="GoodTitle" className="is-vcentered">
-            Jam
-          </td>
-        </tr>
+        <table className="table">
+          <tbody>
+            {goods.map((good, index) => {
+              return (
+                <tr
+                  key={index}
```
zibi-dev, line 61: Using the index as a key is not a good practice. In this case, it doesn’t matter, but remember to avoid it. If you have a unique value, use that instead.

zibi-dev (approved):
Good job! Left some comments for future reference.

